PR #30, opened 2019-10-15
Configuration parser help

### LINKED ISSUE #19: Configuration engine ignores unknown fields
If you misspell a variable name in your config, it will just be silently ignored and a default value will be used instead. 

This is Bad™.

Must be fixed.

sklirg:
This feature aims to resolve #19 by using reflection on the senders, receivers and handlers to identify their properties. During startup, all properties are automatically parsed by Go if everything is configured properly. However, this adds a routine to run during initialization to verify the specified configuration parameters with regards to superfluous fields (which in theory could be mistyped configuration parameters).

Dependencies:

* Adds sirupsen/logrus as a logging framework and uses this throughout the feature. 

Examples:

```json
{
  "receivers": {
    "bar": {
      "type": "udp",
      "address": "[::1]:5015",
      "superfluousField": "sure thing",
      "handler": "protobuf"
    }
  }
}
```

In this configuration there is an excess field. This could be a mistyped field, which potentially could change how things work (especially if it is an optional field), and it will produce a warning, which is shown in the following output:

```
WARN[0000] Configuration property configured but not defined in code (this property won't change anything, is it wrongly defined?)  property=superfluousField
```

### DIFF
--- a/config/parse.go
+++ b/config/parse.go
@@ -211,7 +211,10 @@ func File(f string) (*Config, error) {
 // completion.
 func resolveSenders(c *Config) error {
 	for _, s := range skogul.SenderMap {
+		log.WithField("sender", s.Name).Debug("Resolving sender")
+
 		if c.Senders[s.Name] == nil {
+			log.WithField("sender", s.Name).Error("Unresolved sender reference")
 			return skogul.Error{Source: "config parser", Reason: fmt.Sprintf("Unresolved sender reference %s", s.Name)}
 		}
 		s.S = c.Senders[s.Name].Sender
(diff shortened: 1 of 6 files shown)


config/parse.go
```diff
@@ -291,7 +350,77 @@ func verifyItem(family string, name string, item interface{}) error {
...
+
+	return requiredProps
+}
+
+func getRelevantRawConfigSection(rawConfig *map[string]interface{}, family, section string) map[string]interface{} {
+	// log.Debugf("Fetching config section '%s' for '%s' from %v", section, family, rawConfig)
+	return (*rawConfig)[family].(map[string]interface{})[strings.ToLower(section)].(map[string]interface{})
+}
+
+func findMissingRequiredConfigProps(rawConfig *map[string]interface{}, family, handler string, T reflect.Type) []string {
```
KristianLyng: "handler" should be "section" to match getRelevantRawConfigSection - and because it isn't always a handler.

KristianLyng: Ok, having read this function: Doesn't this essentially make _all_ configurable variables required? Or am I missing something?

sklirg: That seems to be correct, yes. How do you think we should solve this? It could delegate configuration verification to the Verify() function and skip it completely here, only warning about superfluous parameters. Or warning about both, but not aborting the init/startup of skogul.

KristianLyng: I think Verify() is the only thing that knows whether something is required or not. I'm thinking we leave it up to that to complain about missing required configuration. Even warning would be misleading...

receiver/http.go
```diff
@@ -128,19 +132,26 @@ func (htt *HTTP) Start() error {
...
-	}
+
 	server.Addr = htt.Address
 	if htt.Certfile != "" {
-		log.Printf("Starting http receiver with TLS at %s", htt.Address)
+		log.WithField("address", htt.Address).Info("Starting http receiver with TLS")
 		log.Fatal(server.ListenAndServeTLS(htt.Certfile, htt.Keyfile))
 	} else {
-		log.Printf("Starting INSECURE http receiver (no TLS) at %s", htt.Address)
+		log.WithField("address", htt.Address).Debug("Starting INSECURE http receiver (no TLS)")
```
KristianLyng: Arguably, I would invert this: I think starting in insecure mode is more important than starting with TLS :D But that's just the way I like _my_ bikeshed to be painted.

receiver/http.go
```diff
@@ -128,19 +132,26 @@ func (htt *HTTP) Start() error {
...
 		log.Fatal(server.ListenAndServe())
 	}
 	return skogul.Error{Reason: "Shouldn't reach this"}
 }
+
+// Verify verifies the configuration for the HTTP receiver
+func (htt *HTTP) Verify() error {
+	if htt.Address == "" {
+		log.Warn("Missing listen address for http receiver, using Go default")
+	}
```
KristianLyng: This is an interesting route to go down that I didn't consider, but which makes sense. If we do this, we should probably eventually move the config-related log-messages in Start()  down here too (e.g.: with/without TLS). But good idea. (also, will be a small merge conflict, but that's not a problem)

KristianLyng: By "this" I mean: Just warn/log in Verify when things are valid, but potentially breaks POLA.

PR #31, opened 2019-10-15
Add Influxdb MeasurementFromMetadata setting

### LINKED ISSUE #12: Allow influxdb to pick up which measurement to write to from metadata
Today the InfluxDB sender gets its measurement from initialization of the sender, meaning that all data written to that sender will go to the same measurement.

There's no need for this. Instead, we can also offer a configuration variable, e.g. "MeasurementFromMetadata", which, if specified, tells the influxdb sender to read the metadatafield specified and use that as the measurement to write to.

It should probably still have some sort of fallback or default.

KristianLyng:
Fixes #12

The big XXX here isn't really new to influxdb. I'm thinking a log
message is in order, but I've left it out for now because I really think
we need to revamp the logging in general now that I've/we've got a
better understanding of what we really need.

### DIFF
--- a/sender/influxdb.go
+++ b/sender/influxdb.go
@@ -39,12 +39,13 @@ InfluxDB posts data to the provided URL and measurement, using the InfluxDB
 line format over HTTP.
 */
 type InfluxDB struct {
-	URL         string          `doc:"URL to InfluxDB API. Must include write end-point and database to write to." example:"http://[::1]:8086/write?db=foo"`
-	Measurement string          `doc:"Measurement name to write to."`
-	Timeout     skogul.Duration `doc:"HTTP timeout"`
-	client      *http.Client
-	replacer    *strings.Replacer
-	once        sync.Once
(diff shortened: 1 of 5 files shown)


KristianLyng:
@sklirg ping, mind doing a quick ack/no-ack? :)

sender/influxdb.go
```diff
@@ -58,7 +59,26 @@ func (idb *InfluxDB) Send(c *skogul.Container) error {
...
+			// it also catches the scenario where the type cast
+			// is successful, but the key is empty.
+			if measurement == "" {
+				// XXX:
+				// How do we report issues of single
+				// metrics failing, but not the container
+				// in general? Failing the entire container
+				// for just one failed metric is not really
+				// acceptable...
+				continue
```
sklirg: This could be configured by the end user. Is a metric failing transform any relevant? If yes: use it, if no: discard it. This was implemented in the Split transformer. Can be left as a "todo" (new GH issue?) for now.

KristianLyng: Makes a certain amount of sense, I'll open a separate issue on it, because I'm not coming up with a good name for it right this moment.

PR #33, opened 2019-10-16
feat: Initial support for extracting a nested map down to the root

sklirg:


transformer/metadata.go
```diff
@@ -75,11 +75,67 @@ func (meta *Metadata) Transform(c *skogul.Container) error {
...
+
+			nestedObjArray, ok := obj[nestedPath[len(nestedPath)-1]].([]interface{})
+			if !ok {
+				fmt.Println("Failed to cast to []interface{}, aborting transform")
+				return skogul.Error{Reason: "Failed cast"}
+			}
+
+			nestedObj = make(map[string]interface{})
+			for i, val := range nestedObjArray {
+				fmt.Printf("writing arr  i '%s' to new: %d = %+v\n", fmt.Sprintf("%d", i), i, val)
```
KristianLyng: There's a lot of fmt.Printf here. I'm ok with it for now, but:

1. This actually a performance hit. I did a quick test, and doing just Println to /dev/null is 600-650 ns/op alone. By itself that's not a lot, but it adds up. If I'm not much mistaken, this is mainly because it will trigger system calls, witch means context switching into kernel space.
2. We should use log.* instead at the very least. Even without logrus, it makes it obvious that this is something you/we _want_, not just a left-over "work-in-progress" printf. It will also allow filtering in the future, and since we can divert the writes, we can do the handling of logging in a different thread.... I mean go routine.

But: I am not requesting a fix for this now, because getting this merged ASAP is more important. But it makes it obvious we need to prioritize logrus-integration.

KristianLyng: For comparison: json-parsing is 5000ns/op on this system, so a single fmt.Println (to /dev/null, not an actual terminal) is 10% of a parsing of JSON. Container Validation is 3.26ns/op ....

transformer/metadata.go
```diff
@@ -75,11 +75,67 @@ func (meta *Metadata) Transform(c *skogul.Container) error {
...
+
+				obj, isMap := val.(map[string]interface{})
+
+				// If the cast is successful, the array of items is a list of map[string]interface{},
+				// and we want to extract each key to its own key in the root, prefixed with the path
+				// Otherwise, the array is a list of a primitive construct and we
+				// simply prefix the key with the array index
+				if isMap {
+					for key, val := range obj {
+						nestedObj[fmt.Sprintf("%d__%s", i, key)] = val
```
KristianLyng: The separator should probably be configurable. I think a lot of people would like/prefer a dot, for example. Let's make that version 2.

transformer/metadata.go
sklirg: Fixed. Was only for debugging purposes.

PR #44, opened 2019-10-22
http-sender: Support alternate root CA

### LINKED ISSUE #43: HTTP sender: Use alternate alternate root CA
SSIA.

KristianLyng:
Fixes #43

Had to use ht.ok, since init() is just run once. It shouldn't fail,
since Verify will check if it works, but in the scenario where something
changes between startup/Verify and the first metric sent with the
sender, it could potentially fail, so we need to handle it.

Not really any way to prevent that without an API change.


sender/http.go
```diff
@@ -46,13 +48,92 @@ type HTTP struct {
...
+// done to satisfy Verify()'s requirement of not modifying state and it
+// being optional. We also need Verify to actually test this, so the user
+// can be reasonably certain that a valid configuration is used.
+func getCertPool(f string) (*x509.CertPool, error) {
+	// this means "use system default"
+	if f == "" {
+		return nil, nil
+	}
+	cp := x509.NewCertPool()
+	fil, err := os.Open(f)
```
sklirg: I'd prefer if `f` was called `path` and `fil` could be `f`. Minor nitpick, though.

KristianLyng: Makes sense. I'll fix it before I merge.

PR #48, opened 2019-10-23
feat: Generate protobuf go files during go generate

### LINKED ISSUE #34: Use go:generate or similar to generate/update telemetry protobuf code
The gen/ code is generated with a protobuf compiler and Juniper Telemetry data. How this is done should be set up in 'go:generate' or similar so it can be reproduced/updated.

sklirg:
Resolves #34 to generate go files from the protobuf definitions. This PR does not include the protobuf sources for the currently generated files as per the question in a comment in #34.

### DIFF
--- a/gen.go
+++ b/gen.go
@@ -1,0 +1,3 @@
+package skogul
+
+//go:generate /bin/bash -c "prototool generate || echo '❌ Could not generate go files from prototcol buffer definitions'"


sklirg:
Fixed in 404ceea9e7faccdb48708e6de9a1b9bc5a620e17

Issue #39, opened 2019-10-21
Migrate the code from log to logrus


KristianLyng:
Logrus has been introduced, but we need to migrate all code to it.

I think we also need a discussion/policy on how/when to log at all.

In general, I'm thinking:

1. Code should not use log.Fatal unless it's package main: it prevents trivial testing.
2. Successful code paths should not log. Exceptions can be made for important debugging, but for things like transformers, intermediary senders, etc, I'd _prefer_ if it didn't log on OK code paths. It is a performance hit...
3. If we return error, logging should be "trace" at most - the calling code should log the error at higher log levels instead (e.g.: if HTTP sender fails, log trace, and throw error, that way, the retry sender can decide what to log).
4. If we discard actual errors: Log with at least info if it is a common setup. Preferably higher level. 
5. If we handle errors (e.g.: fallback-sender), log with debug.

I think this list needs to be improved/revised as we see how the logging works in real life, but some sort of common principles are needed.

KristianLyng:
I think we've covered this pretty well, but there are two points missing right now that I'm aware of:

[ ] I'll write a small document for docs/ to try to formulate how we approach logging. It shouldn't really be all that complex.
[ ] I re-introduced old-style log in the edit/regex/replacement transformer. I'll fix that.

Since this is mostly my mess, I'll take it from here :D

KristianLyng:
So #61 solved most of the code-issues, and I made #64 to follow up this, it's not as critical as the rest of the task and we could use some time to see the results of #64 and logrus in general.

PR #49, opened 2019-10-23
Migrate to logrus

sklirg:
This PR does not necessarily address the points listed in #39.

- [ ] 1. Code should not use log.Fatal unless it's package main: it prevents trivial testing.
    - This has not been changed; only been ported.

- [x] 2. Successful code paths should not log. Exceptions can be made for important debugging, but for things like transformers, intermediary senders, etc, I'd prefer if it didn't log on OK code paths. It is a performance hit...
    - This is the norm as of now for performance-sensitive code. Some 'OK' paths are logged during init/setup, which I think is fine.

- [ ] 3. If we return error, logging should be "trace" at most - the calling code should log the error at higher log levels instead (e.g.: if HTTP sender fails, log trace, and throw error, that way, the retry sender can decide what to log).
    - Same as ^1: Nothing has been changed with regards to this. 

- [ ] 4. If we discard actual errors: Log with at least info if it is a common setup. Preferably higher level.
    - Not looked into

- [ ] 5. If we handle errors (e.g.: fallback-sender), log with debug.
    - Not looked into

---

I think this can be merged and the unresolved points can remain up for discussion/fixing. Rather get logrus in as the default logger now and fix how runtime should work later.


KristianLyng:
I think this looks pretty good, I have no qualms with using this as a starting point. I didn't really expect/want a "fix" for all the points I raised in #39, that was more of a list probably suited for a file in docs/HACKING.rst or something like that. 

IMO we should merge this.

PR #51, opened 2019-10-24
Update prototool config to work for new Juniper proto definitions

sklirg:
* Excludes the protobuf definitions which has colliding extension numbers
* Update the path to where the protobuf definitions are


sklirg:
I'll merge this just because it makes the prototool config in-line with how the current generate script works. We'll can remove the config altogether later on if we decide against using prototool.

Issue #38, opened 2019-10-21
Transform content of metrics/metadata keys


KristianLyng:
E.g. Juniper sends systemId of the form "hostname-reX:IP", and we want to extract "hostname" from it. And the sensorName is typicially "foo:/foo-ish:/foo/foo/foo:PFE" where we just want "foo".

I'd like some benchmarks if we're doing a general purpose regex-thing, but it should be fine.

A transformer should probably:

1. Be able to read and write from the same variable (e.g.: systemId ~= s/-re.*$//)
2. Be able to read from a different variable than it writes to (E.g.: keep the original. systemIP = systemId; systemIP ~= s/^.*://)

Maybe something like: 
```
  "type": "regex",
  "source": "systemId",
  "target":  "hostname",
  "find": "^[^-]*"
```

or similar?

We probably want both "replace" and "find"?

KristianLyng:
Fixed with #55

PR #55, opened 2019-11-01
Feature/issue 38 transform content of metrics

KristianLyng:
This should cover most of what #38 is really about. The corner-case of non-string'able source fields isn't perfect, but I don't really see any valid way of handling this.

### DIFF
--- a/transformer/edit.go
+++ b/transformer/edit.go
@@ -32,7 +32,7 @@ import (
 	"github.com/telenornms/skogul"
 )
 
-// Replace executes a set of regular expressions, in order
+// Replace executes a regular expression replacement of metric metadata.
 type Replace struct {
 	Source      string `doc:"Metadata key to read from."`
 	Destination string `doc:"Metadata key to write to. Defaults to overwriting the source-key if left blank. Destination key will always be overwritten, e.g., even if the source key is missing, the key located at the destination will be removed."`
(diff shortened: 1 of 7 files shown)


transformer/edit.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"fmt"
+	"log"
+	"regexp"
+	"sync"
+
+	"github.com/telenornms/skogul"
+)
+
+// Replace executes a set of regular expressions, in order
```
sklirg: Minor nitpick: I can't see the possibility of supplying multiple regular expressions.

Unless you are thinking of if you add multiple regexes to the same expression. If that's the case, I think it's clear from the supplied regex how it will be executed.

Should the docstring rather be "Replace executes a regular expression" ?

transformer/edit.go
```diff
@@ -0,0 +1,104 @@
...
+	replace.once.Do(func() {
+		if replace.Destination == "" {
+			replace.Destination = replace.Source
+		}
+		replace.regex, replace.err = regexp.Compile(replace.Regex)
+	})
+	// Verify() should catch this, so there's no reasonable way this
+	// should happen. But in the off chance that a regex compiles on
+	// the first attempt but not the second.... (e.g.: some serious
+	// bugs)
```
sklirg: I don't think we need this. :p Especially if the verify should pick up a compile error of the regex, I think we should trust the underlying implementation enough to work consistently 😅

transformer/edit.go
KristianLyng: You're right, it should. I wrote it before I finished the implementation.

transformer/edit.go
KristianLyng: But this assumes we run Verify() under all thinkable scenarios, though. I'm somehow less concerned about regexp failing to compile on the second run than us somehow failing to do the Verify() under some corner case in the future... And I figure asserts are never really wrong...

Admittedly, I wrote this BEFORE I wrote Verify(), but still....

PR #61, opened 2019-11-05
Introduce skogul.Logger() and use it

KristianLyng:
I'm not 100% sold on this idea, but: If we are going down the logrus-route, we need to ensure the data fields we provide are usable. That means some sort of theme or standardized minimum.

What bugs me about this, and logrus for that matter, is that the code becomes cluttered. I much prefer log.Printf("foo: %v", x) over this WithField("foo",x) thing. This patch also means each code-entity needs a local "logger-instance", which isn't a big deal, but means that log.Print("HI") becomes localLog.Print("HI"). 

This patch mainly aims to unify our approach to logging, and can be a starting point. It is best read by reading common.go and just a couple of other of the files, since it's mostly identical fixes to introduce "local scope". The result is that every log item will have a field "source" referencing what it is (sender/receiver/transformer/parser/core) and a separate field indicating the implementation. E.g.:

```
INFO[0001] Starting INSECURE http receiver (no TLS)      address="[::1]:1386" receiver=http source=receiver
ERRO[0001] HTTP response status code was not in OK range  error="http sender: non-OK status code from target: 400" sender=http source=sender
ERRO[0001] Failed to do POST request                     error="http sender: unable to POST request: Post http://[::1]:1/foobar: dial tcp [::1]:1: connect: connection refused" sender=http source=sender
INFO[0001] unable to read root ca: EOF                   sender=http source=sender
ERRO[0001] Invalid item configuration                    core=config family=sender name=bad1 source=core
INFO[0001] unable to open alternate root CA: open /dev/no/proc/such/sys/file/run/lol/kek/why/are/you/still/reading/this: no such file or directory  sender=http source=sender
ERRO[0001] Invalid item configuration                    core=config family=sender name=bad2 source=core
DEBU[0002] Verifying sender configuration                core=config sender=bad source=core
ERRO[0002] Invalid item configuration                    core=config family=sender name=bad source=core
DEBU[0002] Verifying sender configuration                core=config sender=bad source=core
ERRO[0002] Invalid item configuration                    core=config family=sender name=bad source=core
DEBU[0002] Resolving sender                              core=config sender=fail source=core
DEBU[0002] Resolving sender                              core=config sender=ok source=core
DEBU[0002] Resolving sender                              core=config sender=ignore source=core
DEBU[0002] Resolving sender                              core=config sender=fake-fail source=core
DEBU[0002] Verifying sender configuration                core=config sender=ignore source=core
```

I think we can discuss this, and merge it to try it out, even if we reserve the right to reverse our approach to logging.

References #39

### DIFF
--- a/common.go
+++ b/common.go
@@ -25,6 +25,7 @@ package skogul
 
 import (
 	"fmt"
+	"github.com/sirupsen/logrus"
 	"runtime"
 	"time"
 )
@@ -221,6 +222,14 @@ func (h Handler) Verify() error {
 		return Error{Reason: "Missing parser for Handler"}
 	}
 	return nil
+}
+
+// Logger returns a logrus.Entry pre-populated with standard Skogul fields.
+// sourceType is the typical family of the code/module:
+// sender/receiver/parser/transformer/core, while sourceName is the local
+// implementation.
+func Logger(sourceType, sourceName string) *logrus.Entry {
+	return logrus.WithField("source", sourceType).WithField(sourceType, sourceName)
 }
 
 // AssertErrors counts the number of assert errors
--- a/config/parse.go
+++ b/config/parse.go
@@ -42,6 +42,8 @@ import (
 	"github.com/telenornms/skogul/transformer"
 )
 
+var confLog = skogul.Logger("core", "config")
+
 // Sender wraps the skogul.Sender for configuration parsing.
 type Sender struct {
 	Type   string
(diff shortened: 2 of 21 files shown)


sklirg:
Regarding the following concern:

> What bugs me about this, and logrus for that matter, is that the code becomes cluttered. I much prefer log.Printf("foo: %v", x) over this WithField("foo",x) thing.

This comes from the approach to structured logging. If we want to be able to easily search through logs, having dedicated fields to query makes it easier to find what we are looking for. I added the WithField(s) statements with the assumption that the logs would be passed to an indexing service at some point. If that's not the case we can be more lenient regarding the log formatting/"fielding".

Logrus also has some configuration options, e.g. log things as json (for structured logs), which we could look at.

KristianLyng:
I agree to everything.... I'm merging it and we can continue tweaking it later.

PR #62, opened 2019-11-05
config: Account for tabs in syntax errors

KristianLyng:
Still doesn't account for other non-1-char-wide characters though, but
this should cover 99% of all related issues.


KristianLyng:
Ok this is such a no-brainer that I'm merging it with 0 review now.

PR #69, opened 2019-11-11
Add telemetry.rst and CONTRIBUTING docs

KristianLyng:
I'm guessing both can be improved. Some feedback would be useful.


KristianLyng:
Hoping for a review from @lasseh as well.

PR #71, opened 2019-11-11
feat: Conditional transformation

### LINKED ISSUE #50: Conditional transformation based on a field in the metadata
### Feature request

I propose being able to apply a transformer to an incoming metric based on a conditional, e.g. matching it against a metadata field (sensorName, systemId, ...).

### Rationale

It is possible to apply multiple transformers to the same incoming container/metric, e.g. to first split it and then flatten it.

All transformers are being tried before the container is handed over to the sender.

Currently, we have configured 10 transformers where approximately 3 transformers are run per metric (split, extract, flatten). The other ones silently fail, e.g. because the target field for extract or flatten doesn't exist. I think this is "abusing" the way transformers work.

I find two potential pain points of this. 

1) The configuration can be hard to maintain simply because of the number of transformers applied to the same handler.
2) A simple transformer might apply to multiple metrics which could end up transforming/mangling the metrics in some way the user doesn't want or expect it to do, and there is no way around this (it's nice that it _can_ happen for all of the metrics, e.g. extracting the interface name from the data set, but I think it should be possible to skip it as well)

Furthermore, I haven't really dug deep regarding this, but some of the transformers might run for a bit before they fail which could have an impact on performance. E.g. extracting nested json-objects by casting a metric into `map[string]interface{}` on multiple levels before it realises the field it actually wants to extract doesn't actually exist.

### Potential solution

This feature could be implemented in the transformers, but at that point you wouldn't really solve the maintenance of the configuration issue.

I think that extending the handlers with a configuration option for mapping a transformer to a conditional would be the way to go.

Potential configuration:

```json
{
  "receivers": {
    "udp": {
      "type": "udp",
      "address": "[::1]:1337",
      "handler": "next"
    }
  },
  
  "handlers": {
    "next": {
      "parser": "json",
      "transformers": [
        {
          "transformer": "interface_info",
          "when": "sensorName",
          "is": "junos_system_linecard_interface"
        }
      ]
    }
  },
  
  "transformers": {
    "interface_info": {
      "type": "split",
      "field": "if_name"
    }
  }
}
```

sklirg:
Example config to play around with:
```json
{
  "receivers": {
    "udp": {
      "type": "udp",
      "address": "[::1]:5015",
      "handler": "protobuf"
    }
  },
  "handlers": {
    "protobuf": {
      "parser": "protobuf",
      "transformers": ["sw"],
      "sender": "print"
    }
  },
  "transformers": {
    "sw": {
      "type": "switch",
      "cases": [
        {
          "when": "sensorName",
          "is": "junos_system_linecard_intf-exp:/junos/system/linecard/intf-exp/:/junos/system/linecard/intf-exp/:PFE",
          "transformers": ["remove"]
        }
      ]
    },
    "remove": {
      "type": "data",
      "remove": ["interfaceExp_stats"]
    }
  },
  "senders": {
    "print": {
      "type": "debug",
      "prefix": "DEBUG"
    }
  }
}
```
Change the sensorName conditional to match whatever sensor to test against, or swap it out for something completely different.

---

Resolves #50.

### DIFF
--- a/config/parse.go
+++ b/config/parse.go
@@ -310,7 +310,7 @@ func resolveHandlers(c *Config) error {
 			logger = logger.WithField("transformer", t.Name)
 			logger.Debug("Using predefined transformer")
 
-			h.Handler.Transformers = append(h.Handler.Transformers, *t.T)
+			h.Handler.Transformers = append(h.Handler.Transformers, t.T)
 		}
 	}
 	for _, h := range skogul.HandlerMap {
@@ -335,7 +335,7 @@ func resolveTransformers(c *Config) error {
 			return skogul.Error{Source: "config", Reason: fmt.Sprintf("Unknown transformer %s", t.Name)}
 		}
 
-		t.T = &c.Transformers[t.Name].Transformer
+		t.T = c.Transformers[t.Name].Transformer
 	}
 	return nil
 }
--- a/config/parse.go
+++ b/config/parse.go
@@ -323,6 +323,9 @@ func resolveHandlers(c *Config) error {
 	return nil
 }
 
+// resolveTransformers looks in the parsed config for transformers and initializes the
+// actual transformers. Zeroizes the TransformerMap after if case a new
+// config is applied without restarting.
 func resolveTransformers(c *Config) error {
 	logger := confLog.WithField("method", "resolveTransformers")
 	for transformerName, t := range skogul.TransformerMap {
@@ -337,6 +340,7 @@ func resolveTransformers(c *Config) error {
 
 		t.T = c.Transformers[t.Name].Transformer
 	}
+	skogul.TransformerMap = skogul.TransformerMap[0:0]
 	return nil
 }
 
--- a/config/parse.go
+++ b/config/parse.go
@@ -98,7 +98,7 @@ func (r *Receiver) MarshalJSON() ([]byte, error) {
 }
 
 // UnmarshalJSON picks up the type of the Receiver, instantiates a copy of
-// that receiver, than unmarshals the remaining configuration onto that.
+// that receiver, then unmarshals the remaining configuration onto that.
 func (t *Transformer) UnmarshalJSON(b []byte) error {
 	type tType struct {
 		Type string
@@ -122,7 +122,7 @@ func (t *Transformer) UnmarshalJSON(b []byte) error {
 }
 
 // UnmarshalJSON picks up the type of the Receiver, instantiates a copy of
-// that receiver, than unmarshals the remaining configuration onto that.
+// that receiver, then unmarshals the remaining configuration onto that.
 func (r *Receiver) UnmarshalJSON(b []byte) error {
 	type tType struct {
 		Type string
--- a/config/parse.go
+++ b/config/parse.go
@@ -33,6 +33,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/sirupsen/logrus"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/telenornms/skogul"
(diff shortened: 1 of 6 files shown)


config/parse.go
```diff
@@ -311,15 +310,15 @@ func resolveHandlers(c *Config) error {
...
+				nextT = c.Transformers[t.Name].Transformer
+			} else if t.Name == "templater" {
 				logger.Debug("Using templating transformer")
 				nextT = transformer.Templater{}
 			} else {
 				logger.Error("Unknown transformer")
-				return skogul.Error{Source: "config", Reason: fmt.Sprintf("Unknown transformer %s", t)}
+				return skogul.Error{Source: "config", Reason: fmt.Sprintf("Unknown transformer %s", t.Name)}
 			}
 			h.Handler.Transformers = append(h.Handler.Transformers, nextT)
```
KristianLyng: Hmm, interesting.... Ok, let's leave the hmm-interesting out of it for now:

There's no point looking up c.Transformers[t.Name] since we have t.T already. So this entire if-thing from 311 to 323 can be shortened down to:

h.Handler.Transformers = append(h.Handler.Transformers, t.T)

This does introduce a change in feature, since we're dropping the "templater" default. But I think that's the correct thing to do, and if we want implicit objects like that, we should instead formalize it. And I suspect it's already broken because resolveTransformers wont be able to resolve the templater transformer anyway.

config/parse.go
```diff
@@ -334,6 +333,23 @@ func resolveHandlers(c *Config) error {
...
+		logger = logger.WithField("transformer", transformerName)
+
+		if c.Transformers[t.Name] != nil {
+			logger.Debug("Using predefined transformer")
+		} else {
+			logger.Error("Unknown transformer")
+			return skogul.Error{Source: "config", Reason: fmt.Sprintf("Unknown transformer %s", t.Name)}
+		}
+
+		t.T = &c.Transformers[t.Name].Transformer
```
KristianLyng: I don't think you need to dereference this? We need to for handlers, but that's because a handler is a type, not an interface. Not sure though...? Just notice we're not doing it for senders.

transformer/switch.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// Case requires a field ("when") and a value ("is") to match
+// for the set of transformers to run
+type Case struct {
+	When         string                  `doc:"Used as a conditional statement on a field"`
+	Is           string                  `doc:"Used for the specific value (string) of the stated metadata field"`
+	Transformers []skogul.TransformerRef `doc:"The transformers to run when the defined conditional is true"`
+}
+
```
KristianLyng: This is rather interesting. I don't think we expose nested doc strings, but we probably should. At the moment, this probably means that the end-user documentation is a bit lacking, but I'm inclined to accept that, and instead raise a separate issue for config/help.go to detect this and expose nested documentation/custom types. Shouldn't be TOO hard, and I'd rather do that than limit our options in how we code modules.

transformer/switch.go
```diff
@@ -0,0 +1,70 @@
...
+// Switch is a wrapper for a list of cases
+type Switch struct {
+	Cases []Case `doc:"A list of switch cases "`
+}
+
+var logger = skogul.Logger("transformer", "switch")
```
KristianLyng: Keep in mind: This is package scope, so you just declared "logger" for every transformer, which I think will be rather confusing long-term. I suggest switchLogger or similar.

config/parse.go
sklirg: To make sure I've understood you correctly: It's fine to drop support for the templating transformer? And if it's to be used, it should be reimplemented to work more in line with how other transformers work?

transformer/switch.go
sklirg: More docs == better? :D It gives me documentation in my editor, at least, so that's nice.

config/parse.go
KristianLyng: Yeah, you can still specify it in "transformers" today, just "type": "templating" with no other info. So it's not dropping support, it's just bringing it in line with the others, now that there are a whole bunch of them.

sklirg:
Not sure why the PR build started failing now that I fixed the push build (in ec8b3cc)... Hopefully it's a cache problem :)

PR #72, opened 2019-11-12
feat: Make HTTP receiver auth specific to a handler

sklirg:


receiver/http.go
```diff
@@ -80,6 +102,18 @@ func (rcvr receiver) answer(w http.ResponseWriter, r *http.Request, code int, in
...
+		ok, err := rcvr.auth.auth(r)
+		if !ok && err == nil {
+			return 401, skogul.Error{Source: "http receiver", Reason: "Auth error"}
+		}
+
+		if !ok {
+			// !ok could be either "yeah no this didn't authenticate" or "yeah no this doesn't have access"...
+			return 401, err
+		}
+	}
```
KristianLyng: Suggestion: Make auth() accept the nil-value. E.g., remove alle the above lines so it only reads: if err := rcvr.auth.auth(r); err != nil { return 401, ....} .... Then have auth() do: if auth == nil { return nil }? It underlines the fact that there is no scenario where you get an error but that is ok, and makes handle() trivial.

receiver/http.go
```diff
@@ -57,12 +61,30 @@ type HTTP struct {
...
 
 type httpReturn struct {
 	Message string
 }
 
+func (auth *HTTPAuth) auth(r *http.Request) (bool, error) {
+	var authErrMsg skogul.Error
+
+	// If username is present we assume username/pw auth
+	if auth.Username != "" {
```
KristianLyng: Go best-practice would have you invert this to "if auth.Username == "" { return false, nil }". Not a huge deal.

On a different note: as long as auth only has username/password, we should ensure that either none or both are specified. In the existing implementation this is done on startup, but should be done in Verify() in the future.

sklirg: My thought process regarding the first point here was to be able to re-use part of this logic for the upcoming client cert validation feature, so I didn't want to return early until all auth checks were exhausted.
Will fix second point.

PR #74, opened 2019-11-13
Generalize modules and expose custom data structures

### LINKED ISSUE #73: Add mechanic for exposing documentation for custom data structures
The HTTP receiver and the Switch transformer both use a custom struct/type as part of the config. But we don't expose the documentation for it.

I propose we add an optional 'custom data structures'-part to the auto-system, so you can inform the auto-system that these items have custom data strctures that it needs to expose. Ideally we shouldn't require this, since it's possible to do this by reflection, but the complexity of doing it through reflection versus just requiring an explicit declaration is such that I don't want to go down that road.

This is a reasonably high priority issue, since we currently have undocumented security-features.

KristianLyng:
The worthwhile changes here moves most of the logic from */auto.go into
modules.go. It's the same basic logic, but instead of
Receiver/Sender/Transformer, there's a generic Module structure.

This dramatically reduces the complexity of config/help.go and
cmd/skogul/main.go.

This also introduces "Extras" to the Module data structure, which allows
modules to list a set of custom data structures that should also be
visible in documentation. The Extras field is currently _only_ used for
documentation.

For review, begin with modules.go, comparing it to e.g. sender/auto.go.

Fixes #73

### DIFF
--- a/modules.go
+++ b/modules.go
@@ -1,0 +1,62 @@
+/*
+ * skogul, module automation utilities
+ *
+ * Copyright (c) 2019 Telenor Norge AS
+ * Author(s):
+ *  - Kristian Lyngstøl <kly@kly.no>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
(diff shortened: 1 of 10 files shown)


modules.go
```diff
@@ -0,0 +1,62 @@
...
+// Module is metadata for a skogul module. It is used by the receiver,
+// sender and transformer package. The Alloc() function must return a
+// data structure that implements the relevant module interface, which is
+// checked primarily in config.Parse.
+//
+// See */auto.go for how to utilize this, and config/help.go,
+// cmd/skogul/main.go for how to extract information/help, and ultimately
+// config/parse.go for how it is applied.
+type Module struct {
+	Name    string             // short name of the module (e.g: "http"
```
sklirg: Here, it seems like you dropped one of these: `)`

modules.go
```diff
@@ -0,0 +1,62 @@
...
+	Aliases []string           // optional aliases (e.g. "https")
+	Alloc   func() interface{} // allocation of a blank module structure
```
sklirg: This is the only thing I'm a bit reluctant about.. Losing out on a bit of type safety/"clarity". But I guess that's a price to pay when missing generics 😅

KristianLyng: Yeah, I know. We could conceivably have type safety with reflection, but I don't think it's worth it. We still get implicit type safety because the moment you try to load a module it is cast and checked, so it's impossible to use (or test) the wrong type.

Issue #64, opened 2019-11-06
Document logging approach


KristianLyng:
Add docs/ for logging. This mostly a follow-up of #49

KristianLyng:
Fixed in #75

Issue #57, opened 2019-11-01
Document architectural guide lines


KristianLyng:
A document listing certain pricniples of the design is in order. It should include such things as:

1. Senders are not allowed to modify metrics. Ever.
2. If a sender creates a new container, that container should never be sent to a new sender direct, but to a handler.
3. A transformer should not set up multiple references to the same object/key (E.g.: copy the content, not a reference).
4. Receivers should, by default, use multiple go routines where possible.

And so on. This is not a comprehensive list, but the document should list these "rules" so they are easier to discuss and follow. It will also mean that policy/rules can be discussed in pull requests in a more transparent way.

KristianLyng:
I think #75 covers the heart of this issue. It should be a living document.

PR #75, opened 2019-11-14
Documentation refresh

KristianLyng:
This is mostly a matter of moving documentation into docs/, but the most
review-worthy bits are docs/CODING.rst which I consider mostly a
starting point where I wrote down my thoughts.

I think it's important that even if we don't find (or do find, and
correct) issues with these guidelines, they can still be changed when we
try to put them into practice. They shouldn't need to be 100% perfect
from day 1.

Other content-changes are mostly trivial adjustments.


KristianLyng:
Fixes #57 and fixes #64 (or at least begins that work)

PR #76, opened 2019-11-14
fix: Append a provisioned transformer; not a nil one

sklirg:
After #71, this piece of code was not updated to use allocated transformers, but rather the transformer from the list it looped over. In some rare occurrences, this could be a nil value.

See example config which fails on 3dfe4ee5abb936add0bc45dc212b06ae9aa0aeff (and newer) in the supplied test case.

### DIFF
--- a/config/parse.go
+++ b/config/parse.go
@@ -58,7 +58,7 @@ type Receiver struct {
 // Handler wraps skogul.Handler for configuration parsing.
 type Handler struct {
 	Parser       string
-	Transformers []skogul.TransformerRef
+	Transformers []*skogul.TransformerRef
 	Sender       skogul.SenderRef
 	Handler      skogul.Handler `json:"-"`
 }
@@ -246,6 +246,9 @@ func printSyntaxError(b []byte, offset int, text string) {
 // references and does a final validation.
 func Bytes(b []byte) (*Config, error) {
 	var jsonData map[string]interface{}
+	skogul.HandlerMap = skogul.HandlerMap[0:0]
+	skogul.SenderMap = skogul.SenderMap[0:0]
+	skogul.TransformerMap = skogul.TransformerMap[0:0]
 	if err := json.Unmarshal(b, &jsonData); err != nil {
 		jerr, ok := err.(*json.SyntaxError)
 		if ok {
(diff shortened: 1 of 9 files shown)


config/parse.go
```diff
@@ -316,7 +316,7 @@ func resolveHandlers(c *Config) error {
 			logger = logger.WithField("transformer", t.Name)
 			logger.Debug("Using predefined transformer")
 
-			h.Handler.Transformers = append(h.Handler.Transformers, t.T)
+			h.Handler.Transformers = append(h.Handler.Transformers, c.Transformers[t.Name].Transformer)
```
KristianLyng: This isn't really a fix.

resolveHandlers() run after resolveTransformers(). There are no valid scenarios where t.T should be unset at this stage if c.Transformers[t.Name].Transformer exists. Setting that is the sole purpose of resolveTransformers() ... so why is this happening in the first place? It means a TransformerRef has been corrupted, which is a much bigger problem than just this specific issue.

hmmm.... soo.....

After about an hour, I think I found the issue.... new commit will sort of explain, but it opens more questions as well.

KristianLyng:
Copy/paste from 6a8f602 : 
So here's the story: A []TransformerRef will be an array of references to transformers. Ok, all good so far.

The problem was this:

Step 1: The TransformerRef is unmarshalled, causing a *pointer* to it to be put on the TransformerMap.
Step 2: The array is appended to. But it needs to be grown. This is done by *copying* the content. Suddenly, the array contain a copy of the original TransformerRef, but TransformerMap points to the original.
Step 3: resolveTransformer() iterates over the TransformerMap, which has no knowledge of the copy of the TransformerRef.
Step 4: resolveHandlers() uses the pre-resolving reference.

This is actually a big problem. We can't track the content of a ResolverRef, only its creation thorugh Unmarshal(). We _might_ solve this by something along the lines of making TransformerRef.T be a
pointer to a Transformer, and then have the TransformerMap track that instead.

For now, this is a correct solution, but it doesn't solve the issue of extremely subtle bugs for module developers who want to do the same thing.... which reminds me I need to go over any code that uses
[]somethingRef.... (Which I did in the next commit)

PR #77, opened 2019-11-18
Feat/35 logrus receiver

sklirg:


receiver/logrus.go
```diff
@@ -0,0 +1,166 @@
...
+	c := skogul.Container{
+		Metrics: []*skogul.Metric{&m},
+	}
+
+	logEntry, err := json.Marshal(c)
+	if err != nil {
+		logrusLogLogger.Error("Failed to marshal logrus log into container")
+	}
+
+	lg.Handler.H.Handle(logEntry)
```
KristianLyng: In this case, there's not really any need to use json marshal/unmarshal any more. You already have the container, and TransformAndSend() exists for this reason. It was done in the original log-receiver because there was no structured data.

So you can remove the entire Parse()-function and skil line 111 through 115 and instead do lg.Handler.H.TransformAndSend(&c). Unless I'm missing something?

sklirg: That makes total sense 👍

PR #84, opened 2019-11-22
Add RPM packages to releases

### LINKED ISSUE #83: Provide RPM package in release
Let's leave the needs of why this could be wanted out of the discussion... Nobody has broken any servers with something an RPM package (might be able) to prevent.

In the future we can probably add support for other package formats as well.

sklirg:
Resolves #83

### DIFF
--- a/skogul.service
+++ b/skogul.service
@@ -6,7 +6,7 @@ Documentation=man:skogul(1) https://github.com/telenornms/skogul
 After=network-online.target
 
 [Service]
-ExecStart=/usr/local/bin/skogul -f /etc/skogul/default.json
+ExecStart=/usr/bin/skogul -show -f /etc/skogul/default.json
 Restart=on-failure
 User=skogul
 Group=skogul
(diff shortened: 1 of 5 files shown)


skogul.service
```diff
@@ -0,0 +1,16 @@
...
+
+[Service]
+ExecStart=/usr/local/bin/skogul -f /etc/skogul/default.json
+Restart=on-failure
+User=skogul
+Group=skogul
+NoNewPrivileges=true
+LimitNOFILE=31337
+ProtectSystem=full
+PrivateTmp=true
```
KristianLyng: Looking at ssh.service as reference, I think we can also add:

ExecStart=/usr/bin/skogul -show -f /etc/skogul/default.json

skogul.service
```diff
@@ -6,7 +6,7 @@ Documentation=man:skogul(1) https://github.com/telenornms/skogul
 After=network-online.target
 
 [Service]
-ExecStart=/usr/local/bin/skogul -f /etc/skogul/default.json
+ExecStart=/usr/bin/skogul -show -f /etc/skogul/default.json
```
KristianLyng: This.... was a bit too quick :D 

Now the path is correct, but instead of having ExecStartPre=... -show and ExecStart=, we only have -show, so it will never start.

PR #87, opened 2019-11-27
Test example configs

sklirg:


docs/examples/http_to_influx.json
```diff
@@ -11,7 +11,7 @@
   "handlers": {
     "jsontemplating": {
       "parser": "json",
-      "transformers": [ "templater" ],
+      "transformers": [],
```
KristianLyng: Considering the name, if we remove the templater transformer, we should rename the handler as well.

Could also just specify a templater transformer.

sklirg:
Makes sense, I added the templating transformer to some configs again :)

PR #90, opened 2019-12-02
feat: Add JSON parser for custom JSON formats 

sklirg:
Example skogul config:

```js
{
    "receivers": {
        "readfile": {
            "type": "file",
            "file": "./mqtt-data.json",
            "handler": "mqtt-parse"
        }
    },
    "handlers": {
        "mqtt-parse": {
            "parser": "custom-json",
            "transformers": ["split-events", "flatten-attributes", "remove-attributes", "extract_sysname"],
            "sender": "print"
        }
    },
    "transformers": {
        "split-events": {
            "type": "split",
            "field": ["Events"]
        },
        "flatten-attributes": {
            "type": "data",
            "flatten": [
                ["attributes"]
            ]
        },
        "remove-attributes": {
            "type": "data",
            "remove": ["attributes"]
            
        },
        "extract_sysname": {
            "type": "metadata",
            "extractfromdata": ["source"]
        }
    },
    "senders": {
        "print": {
            "type": "debug",
            "prefix": ""
        }
    }
}
```

will parse the following data 

```js
{
    "Events": [
        {
            "source": "Devices.Device.wl0p0",
            "type": "network",
            "attributes": {
                "wl0p0": {
                    "Timestamp": "2019-12-02T13:37:22Z",
                    "RxPackets": 0,
                    "TxPackets": 5234,
                    "RxBytes": 0,
                    "TxBytes": 443032,
                    "DeltaRxPackets": 0,
                    "DeltaTxPackets": 3,
                    "DeltaRxBytes": 0,
                    "DeltaTxBytes": 147
                }
            }
        }
    ]
}
```

into a skogul Container like so
```js
{
  "metrics": [
    {
      "timestamp": "2019-12-02T14:37:00.466143+01:00",
      "metadata": {
        "source": "Devices.Device.wl0p0"
      },
      "data": {
        "attributes__wl0p0": {
          "DeltaRxBytes": 0,
          "DeltaRxPackets": 0,
          "DeltaTxBytes": 147,
          "DeltaTxPackets": 3,
          "RxBytes": 0,
          "RxPackets": 0,
          "Timestamp": "2019-12-02T13:37:22Z",
          "TxBytes": 443032,
          "TxPackets": 5234
        },
        "type": "network"
      }
    }
  ]
}
```

### DIFF
--- a/config/parse.go
+++ b/config/parse.go
@@ -308,6 +308,9 @@ func resolveHandlers(c *Config) error {
 		if h.Parser == "protobuf" {
 			logger.Debug("Using protobuf parser")
 			h.Handler.Parser = parser.ProtoBuf{}
+		} else if h.Parser == "custom-json" {
+			logger.Debug("Using custom JSON parser")
+			h.Handler.Parser = parser.RawJSON{}
 		} else if h.Parser == "json" || h.Parser == "" {
 			logger.Debug("Using JSON parser")
 			h.Handler.Parser = parser.JSON{}
--- a/parser/json.go
+++ b/parser/json.go
@@ -25,6 +25,7 @@ package parser
 
 import (
 	"encoding/json"
+	"time"
 
 	"github.com/telenornms/skogul"
 )
@@ -37,4 +38,33 @@ func (x JSON) Parse(b []byte) (*skogul.Container, error) {
 	container := skogul.Container{}
 	err := json.Unmarshal(b, &container)
 	return &container, err
+}
+
+// RawJSON can be used when the JSON format does not conform to the final JSON format of skogul,
+// e.g. when it is used as the first step of parsing from a third party source where modifying
+// the source data structure might be hard/impossible
+type RawJSON struct{}
+
+// Parse accepts a byte slice of JSON data and marshals it into an empty skogul.Container
+func (data RawJSON) Parse(b []byte) (*skogul.Container, error) {
+
+	// The Validate() func of a container expects a timestamp to be valid.
+	// Better way to fix?
+	time := time.Now()
+	metric := skogul.Metric{
+		Metadata: make(map[string]interface{}),
+		Time:     &time,
+	}
+
+	err := json.Unmarshal(b, &metric.Data)
+
+	if err != nil {
+		return nil, err
+	}
+
+	container := skogul.Container{
+		Metrics: []*skogul.Metric{&metric},
+	}
+
+	return &container, nil
 }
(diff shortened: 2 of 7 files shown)


parser/json.go
```diff
@@ -38,3 +39,32 @@ func (x JSON) Parse(b []byte) (*skogul.Container, error) {
...
+// e.g. when it is used as the first step of parsing from a third party source where modifying
+// the source data structure might be hard/impossible
+type RawJSON struct{}
+
+// Parse accepts a byte slice of JSON data and marshals it into an empty skogul.Container
+func (data RawJSON) Parse(b []byte) (*skogul.Container, error) {
+
+	// The Validate() func of a container expects a timestamp to be valid.
+	// Better way to fix?
+	time := time.Now()
```
sklirg: Comment says it all really.

sklirg: In other words, though: 
* Should the parser support extracting a timestamp?
* Should a transformer support being able to override the timestamp? (Should a transformer actually be allowed to override the timestamp?)

KristianLyng: I'm thinking: 
1. For this example, the only valid timestamp _is_ now. It's a custom metric, it might not even have a timestamp.
2. Transformers should most certainly be able to override the timestamp. There are numerous examples where that's useful.
3. We should not do time.Now() for every received metric. We are looking at receiving many thousands of metrics per second and doing an extra systemcall for each one is wasteful. I suggest we create a skogul.Nowish() function and add a timer in in init() in common.go that updates a global timestamp, and Nowish() returns it. I think 1/10th second resolution is acceptable for the scenarios where we don't have more precise data available. But it can always be increased in the future...
4. For this PR, just using time.Now() is acceptable as long as an issue to fix it is also created.

config/parse.go
```diff
@@ -308,6 +308,9 @@ func resolveHandlers(c *Config) error {
 		if h.Parser == "protobuf" {
 			logger.Debug("Using protobuf parser")
 			h.Handler.Parser = parser.ProtoBuf{}
+		} else if h.Parser == "custom-json" {
```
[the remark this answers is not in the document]
KristianLyng: Yeah.... It's probably just an other type of module. Then we can just check parser.Auto["custom-json"] or similar. We don't really need to expose the module-aspect in configuration of course, since there thus far isn't any configuration for any of the parsers. But I'm fine with this for now. It's not exactly super-important yet. But I _do_ predict more parsers in the future.

PR #92, opened 2019-12-03
Timestamp transformer

### LINKED ISSUE #91: Transformer for setting timestamp for a container
There should exist a transformer which supports setting the timestamp of a container. 

As per the following comment: https://github.com/telenornms/skogul/pull/90#discussion_r353153056

sklirg:
Resolves #91.


transformer/timestamp.go
```diff
@@ -0,0 +1,94 @@
...
+		c.Metrics[i].Time = &time
+	}
+	return nil
+}
+
+// parseTimestamp parses a timestamp format name into a timestamp format
+// e.g. rfc3339 will be returned as "2006-01-02T15:04:05Z07:00"
+func parseTimestamp(format string) string {
+	switch format {
+	case "rfc3339", "iso8601": // 🙈
```
KristianLyng: Should support lower and upper case.

transformer/timestamp.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
+// Verify will make sure the required fields are set, or set some sane defaults
+func (config *Timestamp) Verify() error {
+	if config.Source == nil {
+		return skogul.Error{Reason: "Missing source field for timestamp transformer", Source: "timestamp transformer"}
+	}
+	if config.Format == "" {
+		config.Format = time.RFC3339
```
KristianLyng: This should be explicit. I suggest moving it to timeParse() and logging a warning.

transformer/timestamp.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// parseTimestamp parses a timestamp format name into a timestamp format
+// e.g. rfc3339 will be returned as "2006-01-02T15:04:05Z07:00"
+func parseTimestamp(format string) string {
+	switch format {
+	case "rfc3339", "iso8601": // 🙈
+		return time.RFC3339
+	default:
+		return format
```
KristianLyng: This should also trigger a log message to avoid surprises if people type "RCF3339" for example.

transformer/timestamp.go
```diff
@@ -0,0 +1,94 @@
...
+		timestamp, ok := obj[config.Source[len(config.Source)-1]].(string)
+
+		if !ok {
+			timestampLogger.Error("Failed to cast timestamp field to a string")
+			if config.Fail {
+				return skogul.Error{Reason: "Failed to cast timestamp field to a string"}
+			}
+		}
+
+		format := parseTimestamp(config.Format)
```
KristianLyng: sync.Once (.do) should be used. Just store it on config.once and be done with it, once and for all! :dagger:

PR #97, opened 2019-12-19
feat: Add MQTT topic for received message as _mqtt_topic in the payload

### LINKED ISSUE #95: Expose MQTT topic on received message
It would be useful to expose what the topic for an MQTT message is, especially when the subscription consists of wildcards.

sklirg:
Resolves #95 

I'm a bit curious of how unmarshalling and marshalling the incoming json will impact performance. I'll try to set up some benchmarks.

### DIFF
--- a/receiver/mqtt.go
+++ b/receiver/mqtt.go
@@ -48,11 +48,29 @@ type MQTT struct {
 	mc skmqtt.MQTT
 }
 
+func appendTopic(container *skogul.Container, topic string) {
+	for _, metric := range container.Metrics {
+		if metric.Metadata == nil {
+			metric.Metadata = make(map[string]interface{})
+		}
+		metric.Metadata["_mqtt_topic"] = topic
+	}
+}
+
 // Handle a received message.
 func (handler *MQTT) receiver(msg mqtt.Message) {
-	err := handler.Handler.H.Handle(msg.Payload())
+	container, err := handler.Handler.H.Parser.Parse(msg.Payload())
+
 	if err != nil {
-		mqttLog.WithError(err).Error("Unable to handle payload")
+		mqttLog.WithError(err).Error("Failed to parse payload from MQTT message")
+		return
+	}
+
+	appendTopic(container, msg.Topic())
+
+	err = handler.Handler.H.TransformAndSend(container)
+	if err != nil {
+		mqttLog.WithError(err).Error("Error during transform or send container")
 	}
 }
 


sklirg:
Another way to do this is to assume that the incoming data is json (which can be a bad assumption), and modify the bytes as such:
`bytes = []byte(fmt.Sprintf("{\"_topic\": \"%s\",%s", topic, string(bytes)[1:]))`

`go test ./receiver/mqtt_test.go -bench . -benchtime 10s`

...WithTopic1: the current implementation
...WithTopic2: Raw bytes/string rewrite (as shown above)

```
goos: darwin
goarch: amd64
BenchmarkReceiveMQTTMessageWithoutTopic-16    	1000000000	         3.12 ns/op
BenchmarkReceiveMQTTMessageWithTopic1-16      	 6272593	      1919 ns/op
BenchmarkReceiveMQTTMessageWithTopic2-16      	60797860	       199 ns/op
```

I didn't commit the tests because they require the AppendTopic function to be exported, but this is the test file used for the above benchmarks:
<details><summary>click to show</summary>
<p>

```go

package receiver_test

import (
	"encoding/json"
	"testing"

	"github.com/telenornms/skogul/receiver"
)

func TestReceiveMQTTMessageAppendTopic1(t *testing.T) {
	bytes := []byte(`{"data": 42}`)

	newBytes, err := receiver.AppendTopic(bytes, "Topic42", 1)

	if err != nil {
		t.Errorf("Failed to append topic: %v", err)
		return
	}

	var parsed map[string]interface{}
	err = json.Unmarshal(newBytes, &parsed)
	if err != nil {
		t.Errorf("Failed to unmarshal payload after adding topic: %v", err)
		return
	}

	if parsed["_topic"] != "Topic42" {
		t.Error("Topic is not what is expected!")
		return
	}
}

func TestReceiveMQTTMessageAppendTopic2(t *testing.T) {
	bytes := []byte(`{"data": 42}`)

	newBytes, err := receiver.AppendTopic(bytes, "Topic42", 2)

	if err != nil {
		t.Errorf("Failed to append topic: %v", err)
		return
	}

	var parsed map[string]interface{}
	err = json.Unmarshal(newBytes, &parsed)
	if err != nil {
		t.Errorf("Failed to unmarshal payload after adding topic: %v", err)
		return
	}

	if parsed["_topic"] != "Topic42" {
		t.Error("Topic is not what is expected!")
		return
	}
}

func BenchmarkReceiveMQTTMessageWithoutTopic(b *testing.B) {
	bytes := []byte(`{"data": 42}`)

	for i := 0; i < b.N; i++ {
		_, err := receiver.AppendTopic(bytes, "Topic42", 0)

		if err != nil {
			b.Error("Failed to not append topic")
		}
	}
}

func BenchmarkReceiveMQTTMessageWithTopic1(b *testing.B) {
	bytes := []byte(`{"data": 42}`)

	for i := 0; i < b.N; i++ {
		_, err := receiver.AppendTopic(bytes, "Topic42", 1)

		if err != nil {
			b.Error("Failed to append topic")
		}
	}
}

func BenchmarkReceiveMQTTMessageWithTopic2(b *testing.B) {
	bytes := []byte(`{"data": 42}`)

	for i := 0; i < b.N; i++ {
		_, err := receiver.AppendTopic(bytes, "Topic42", 2)

		if err != nil {
			b.Error("Failed to append topic")
		}
	}
}
```

<p></details>

KristianLyng:
Ok, there's a more low-level solution to fiddling with strings here.

We have Handle that does parse, transform and send. Then we have TransformAndSend which skips parsing. But you can also just call ```c,err := h.Parser.Parse(b)``` to parse, then modify the container before using TransformAndSend. The Handle() and TransformAndSend() functions are really just convenience functions.

Looking at the implementation in common.go, I suggest we re-factor it slightly to clean up the API.

Today, a Handler has four methods. Handle, Transform, TransformAndSend and Verify. But the real task of a handler can be split into four:

1. Parse
2. Transform
3. Verify
4. Send

However, there are zero situations I can come up with where you want to issue Send without issuing Transform first and vice versa. And if we expose those as separate methods, we are likely to see bugs in senders in particular where Transform is skipped. And we already have bugs today where Verify() isn't called at all. So I suggest we simplify the API:

```
// Parse a byte string and return a container. Really just a shorthand for h.Parser.Parse()
func (h *Handler) Parse(b []byte) (*Container, error)

// Handle a byte string by parsing it and calling TransformAndSend on the result.
func (h *Handler) Handle(b []byte) error

// TransformAndSend an already parsed container, with an implicit Verify in between.
func (h *Handler) TransformAndSend(c *Container) error
```
For this specific issue of MQTT, that means the pattern in the receiver will be: 

1. h.Parse(...) 
2. modify Container by setting c.Metrics[*].Metadata["_topic"] = topic, 
3. call h.TransformAndSend(). This is a pattern I believe will be fairly common.

However, to reduce the scope of #97, I'll leave it to you if you want to just use h.Parser.Parse(...) => modify => H.TransformAndSend(c), or do the API cleanup. Probably makes sense to keep the tasks separate.

sklirg:
Modified this change request to use Parse and TransformAndSend. I'll create an issue regarding cleaning up the API.

PR #99, opened 2020-01-07
Add config for custom flattening separator and skipping intermediate keys from nested objects

sklirg:


sklirg:
Example:

It is possible to split a metric such as this one
```json
{
  "timestamp": "2020-01-07T13:01:57.74941+01:00",
  "metadata": {
    "foo": "bar"
  },
  "data": {
    "data": {
      "metric0": 6129484611666146000
    },
    "metadata": {
      "key1": 0
    }
  }
}
```

into this
```json
{
  "timestamp": "2020-01-07T13:02:35.126621+01:00",
  "metadata": {
    "foo": "bar"
  },
  "data": {
    "metric0": 6129484611666146000
  }
}
```

using the following transformer configuration:
```json
"transformer": {
  "type": "data",
  "flatten": [["data"]],
  "flattenSeparator": "drop",
  "remove": ["metadata", "data"]
}
```

transformer/metadata.go
```diff
@@ -79,12 +79,20 @@ func (meta *Metadata) Transform(c *skogul.Container) error {
...
+	} else {
+		pathFmt = "%s" + separator + "%s"
+	}
 
-	if len(nestedPath) > 1 {
+	// Create a nested path unless configuration says not to
+	if separator != "drop" && len(nestedPath) > 1 {
 		for _, p := range nestedPath[1:] {
-			newPath = fmt.Sprintf("%s__%s", newPath, p)
+			nestedObjectPath = fmt.Sprintf(pathFmt, nestedObjectPath, p)
```
KristianLyng: Clearly this works, but wouldn't it be cleaner to just use Sprintf("%s%s%s","nestedObjectPath, separator, p) ?

Granted, it doesn't give us _much_, though...

Hmm, what we could do, is make the Data.FlattenSeparator a string *pointer*, instead of a string. It's a bit... weird... but it will allow us to distinguish between a config with a blank separator (valid pointer to an empty string) and a config with an undefined separator (nil). But, for now, I think using "drop" is fine - providing a mechanism for default values for strings is something that belongs in the config parser, not a random module.

PR #108, opened 2020-01-29
Refactor internal handler api

### LINKED ISSUE #98: Refactor internal Handler API
As suggested in https://github.com/telenornms/skogul/pull/97#issuecomment-570158408:

Generally, refactor Handlers to provide the following functions:

```
// Parse a byte string and return a container. Really just a shorthand for h.Parser.Parse()
func (h *Handler) Parse(b []byte) (*Container, error)

// Handle a byte string by parsing it and calling TransformAndSend on the result.
func (h *Handler) Handle(b []byte) error

// TransformAndSend an already parsed container, with an implicit Verify in between.
func (h *Handler) TransformAndSend(c *Container) error
```

The actions to be done in these functions are:

1. Parse
2. Transform
3. Verify
4. Send

sklirg:
Fixes #98

### DIFF
--- a/common.go
+++ b/common.go
@@ -194,13 +194,13 @@ func (h *Handler) Parse(b []byte) (*Container, error) {
 
 // Transform validates the parsed container and runs all available transformers
 func (h *Handler) Transform(c *Container) error {
-	if err := c.Validate(); err != nil {
-		return Error{Source: "handler", Reason: "validating metrics failed", Next: err}
-	}
 	for _, t := range h.Transformers {
 		if err := t.Transform(c); err != nil {
 			return err
 		}
+	}
+	if err := c.Validate(); err != nil {
+		return Error{Source: "handler", Reason: "validating metrics failed", Next: err}
 	}
 	return nil
 }
--- a/common.go
+++ b/common.go
@@ -233,7 +233,7 @@ func (h *Handler) TransformAndSend(c *Container) error {
 // Verify the basic integrity of a handler. Quite shallow.
 func (h Handler) Verify() error {
 	if h.parser == nil {
-		return Error{Reason: "Missing parser for Handler"}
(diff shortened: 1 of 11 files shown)


common.go
```diff
@@ -171,8 +171,20 @@ func (e Error) Container() Container {
...
+	c, err := h.Parser.Parse(b)
+	if err != nil {
+		return nil, Error{Source: "handler", Reason: "parsing data failed", Next: err}
+	}
+	return c, nil
+}
+
+// Transform validates the parsed container and runs all available transformers
 func (h *Handler) Transform(c *Container) error {
+	if err := c.Validate(); err != nil {
```
sklirg: Maybe this should be run as part of Parse() instead of Transform()?

sklirg: Test case added :)

sklirg: Also, fixed.

common.go
```diff
@@ -204,29 +230,26 @@ func (h *Handler) Handle(b []byte) error {
...
 
 // Verify the basic integrity of a handler. Quite shallow.
 func (h Handler) Verify() error {
-	if h.Parser == nil {
-		return Error{Reason: "Missing parser for Handler"}
-	}
-	if h.Transformers == nil {
-		return Error{Reason: "Missing parser for Handler"}
+	if h.parser == nil {
+		return Error{Source: "handler verification", Reason: "Missing parser for Handler"}
```
KristianLyng: At first glance, this one was problematic to me, because, e.g., the logrus receiver doesn't _use_ a parser. Then I realized we already had this test, so obviously it worked.... It turns out we default the parser to "json" if one isn't set in configuration. This makes a lot of sense in general. Most of the time, that's what we want, and for stuff like the logrus receiver, it doesn't matter what the parser is.

So this is actually fine. It will catch corrupt or bugged handlers. Which is what we want anyway.

PR #109, opened 2020-01-29
Make Handler.Parser internal to its module as its functions are exposed already

sklirg:
This makes it easier to avoid "bypassing" the Parse() function exposed by the Handler.

I'm not completely sure if I want to do this or not, so feel free to comment.

### DIFF
--- a/common.go
+++ b/common.go
@@ -44,7 +44,7 @@ used in the future.
 To make it configurable, a HandlerRef should be used.
 */
 type Handler struct {
-	Parser       Parser
+	parser       Parser
 	Transformers []Transformer
 	Sender       Sender
 }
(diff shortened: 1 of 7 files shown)


KristianLyng:
I'm honestly unsure about this myself. I keep going back and forth on it. On the one side, it does enforce using the handlers Parse, and all safeguards that introduce, but it also makes test cases and such more ugly....

On the other hand, looking at the actual changes, there aren't really that many. Most of the time, the handler is contained in the config package, so maybe this is a good thing in general?

It DOES make it clear what the desired usage is... after all, "there's more than one way to do it" is a warning, not a desired state. So I say we go for it.

sklirg:
:shipit:

PR #117, opened 2020-02-13
Run skogul one-off on some data

sklirg:
Adds the `-test` flag which adds a "_debugger" receiver, handler and sender. These can be configured in a configuration file as normal (though the name is required to be "_debugger") to customise, but will default to reading from stdin, parsing as skogul-json, transforming nothing and finally outputting to stdout.

### DIFF
--- a/cmd/skogul/main.go
+++ b/cmd/skogul/main.go
@@ -42,6 +42,8 @@ import (
 	"github.com/telenornms/skogul/transformer"
 )
 
+var mainLogger = skogul.Logger("main", "main")
+
 // versionNo gets set by passing the -X flag to ld like this
 // go build -ldflags "-X main.versionNo=0.1.0" ./cmd/skogul
 var versionNo string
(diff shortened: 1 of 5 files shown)


sklirg:
I'm not completely set on the -test param, but I feel that -one-off is a bit clunky.

cmd/skogul/main.go
```diff
@@ -743,9 +746,17 @@ func main() {
...
 
-	c, err := config.File(*ffile)
-	if err != nil {
-		log.Fatal(err)
+	var c *config.Config
+
+	// Allow config file to be missing if running in -test mode,
+	// require it otherwise.
+	_, confFileErr := os.Stat(*ffile)
+	if !*ftest && os.IsNotExist(confFileErr) {
```
KristianLyng: Doesn't this mean that we will only try reading the config file if -test is missing AND the configfile is missing?

sklirg:
I'll pass on this for now. It's pretty simple to add a basic config file, and with the flexibility of skogul it'll be hard to find a good default behaviour for a test-mode. Better to let the user decide exactly what they want to happen, IMO. Closing.

PR #118, opened 2020-02-26
Rework build-system using Makefile

KristianLyng:
This is a bit of a big change, build-wise. It introduces a traditional Makefile, with various targets, and cleans up the RPM building.

The goal is to have a unified way to build, regardless of platform. It's still not exactly perfect since there's macro-substitution going on for the rpm spec file generation, but it's much better than it was, imo.

I'm unsure if everything is 100% correct, but I think it's time to get this merged and instead fix the fallout, since testing things like drone build pipelines is tricky to do.

I've also bumped the build box to centos:8 to support in-place building. This is a little weird, since our target is really centos 7. But it works.


Makefile
```diff
@@ -0,0 +1,134 @@
...
+	@echo  ❕Building spec-file
+	@cat $< | sed "s/xxVxx/${GIT_DESCRIBE}/g; s/xxARCHxx/${ARCH}/g; s/xxVERSION_NOxx/${VERSION_NO}/g" > $@
+	@which dpkg >/dev/null 2>&1 && { echo 🆒 Adding debian-workaround for rpm build; sed -i 's/^BuildReq/\#Debian hack, auto-commented out: BuildReq/g' $@; } || true
+
+# Build RPM. The spec has a blank %prep, so it assumes sources are already
+# available. This isn't perfect, since it creates a tight coupling between
+# Makefile and specfile, but it isn't all that bad either, since it allows
+# building with minimal redundant effort, and without having to commit to
+# git.
+rpm: rpm-prep/SPECS/skogul.spec | rpm-prep/BUILDROOT/
```
sklirg: This fails for me on my MacOS machine with 
```
$ make rpm
make: *** No rule to make target `rpm-prep/SPECS', needed by `rpm-prep/SPECS/skogul.spec'.  Stop.
```

After creating the rpm-prep/SPECS folder manually, I get 
```
$ make rpm
make: *** No rule to make target `rpm-prep/BUILDROOT', needed by `rpm'.  Stop.
```

Makefile
```diff
@@ -0,0 +1,134 @@
...
+
+check: test fmtcheck vet
+
+vet:
+	@echo 🔬 Vetting code
+	@go vet ./...
+
+fmtcheck:
+	@echo 🦉 Checking format with gofmt -d -s
+	@if [ "x$$(find -name '*.go' -not -wholename './gen/*' -exec gofmt -d -s {} +)" != "x" ]; then find -name '*.go' -not -wholename './gen/*' -exec gofmt -d -s {} +; exit 1; fi
```
sklirg:
```
$ make fmtcheck
🦉 Checking format with gofmt -d -s
find: illegal option -- n
```

By adding the path to search in (".") directly after `find` (`find . -name ...`) everything works. That's required as per my find (BSD version). Note, there are two statements requiring the change. Both inside the if and in the running code if the conditional is true. It's relevant for `fmtfix` too.

Makefile
KristianLyng: What version of make do you use?

sklirg:
```
$ make --version
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

KristianLyng:
@sklirg care to re-test with the latest two commits? Should fix the issues you have on McOS

Makefile
sklirg: Works better after 238fee6. Although the build dependency hack doesn't get applied since I don't have dpkg either. :'D

Could (should?) we add the hack in case the system _does not have_ yum, instead of _if it has dpkg?_

Makefile
sklirg: I think you forgot to add it for fmtfix. Fmtcheck works.

Makefile
```diff
@@ -0,0 +1,131 @@
...
+	@echo 🔬 Vetting code
+	@go vet ./...
+
+fmtcheck:
+	@echo 🦉 Checking format with gofmt -d -s
+	@if [ "x$$(find . -name '*.go' -not -wholename './gen/*' -exec gofmt -d -s {} +)" != "x" ]; then find . -name '*.go' -not -wholename './gen/*' -exec gofmt -d -s {} +; exit 1; fi
+
+fmtfix:
+	@echo 🎨 Fixing formating
+	@find -name '*.go' -not -wholename './gen/*' -exec gofmt -d -s -w {} +
```
sklirg: This find still errors out with: `find: illegal option -- n`

Makefile
```diff
@@ -0,0 +1,131 @@
...
+	@cd docs; \
+	find . -type f -exec install -D -m 0644 {} ${DESTDIR}${DOCDIR}/{} \;
+	@install -D -m 0644 README.rst LICENSE -t ${DESTDIR}${DOCDIR}/
+
+
+# Any complaints on this macro-substitution without patches and I introduce m4.
+build/redhat-skogul.spec: build/redhat-skogul.spec.in
+	@echo  ❕Building spec-file
+	@cat $< | sed "s/xxVxx/${GIT_DESCRIBE}/g; s/xxARCHxx/${ARCH}/g; s/xxVERSION_NOxx/${VERSION_NO}/g" > $@
+	@which dpkg >/dev/null 2>&1 && { echo 🆒 Adding debian-workaround for rpm build; sed -i 's/^BuildReq/\#Debian hack, auto-commented out: BuildReq/g' $@; } || true
```
sklirg: Should this be a "non yum/rpm" hack instead of a debian specific hack? As in, should the hack be applied if the system does not have yum, rather than if it has dpkg? That might make it work better on my MacOS at least. :shipit:

KristianLyng: Who in the world uses non-debian stacks?

And the problem is: I actually _have_ yum and rpm on my debian-box. But I suppose we can check for /etc/redhat-version, that should be fairly reliable. More so than lsb_release at least.

PR #120, opened 2020-03-02
influx: Flip sanity test around to avoid type-issues

### LINKED ISSUE #119: Influx sender doesn't detect map[string]interface{} as illegal field
```
level=error msg="Failed to send data to influx. Bad response from backend." body="{\"error\":\"partial write: unable to parse 'interfaces_interface,componentId=xxx,sensorName=xxxxx,subComponentId=0,systemId=xxx,name=xxx,measurement=interfaces_interface subinterfaces=map[string]interface {}{\\\"subinterface\\\":[]interface {}{map[string]interface {}{\\\"index\\\":0, \\\"ipv6\\\":map[string]interface {}{\\\"addresses\\\":map[string]interface {}{\\\"address\\\":[]interface {}{map[string]interface {}{\\\"ip\\\":\\\"xxx\\\", \\\"state\\\":map[string]interface {}{\\\"status\\\":\\\"PREFERRED\\\"}}}}, \\\"neighbors\\\":map[string]interface {}{}, \\\"state\\\":map[string]interface {}{\\\"enabled\\\":true, \\\"mtu\\\":xxx}, \\\"unnumbered\\\":map[string]interface {}{\\\"interface_ref\\\":map[string]interface {}{\\\"state\\\":map[string]interface {}{\\\"interface\\\":\\\"\\\", \\\"subinterface\\\":0}}, \\\"state\\\":map[string]interface {}{\\\"enabled\\\":false}}}}}} 1583137755443000000': invalid boolean dropped=0\"}\n" category=sender fields=229 metricsSent=14 sender=influxdb status="400 Bad Request" tags=88
```

Should be a straight forward fix.

KristianLyng:
Should satisfy:

Fixes #119

### DIFF
--- a/sender/influxdb.go
+++ b/sender/influxdb.go
@@ -53,10 +53,29 @@ type InfluxDB struct {
 	once                    sync.Once
 }
 
+// checkVariable verifies that the relevant variable is of a type we can
+// handle.
 func checkVariable(category string, field string, idx string, value interface{}) error {
 	t := reflect.TypeOf(value)
 	k := t.Kind()
-	if k == reflect.Struct || k == reflect.Array || k == reflect.Slice {
+
+	switch k {
+	case reflect.Bool:
+	case reflect.Int:
+	case reflect.Int8:
+	case reflect.Int16:
+	case reflect.Int32:
+	case reflect.Int64:
+	case reflect.Uint:
+	case reflect.Uint8:
+	case reflect.Uint16:
+	case reflect.Uint32:
+	case reflect.Uint64:
+	case reflect.Uintptr:
+	case reflect.Float32:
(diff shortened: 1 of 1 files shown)


sender/influxdb.go
```diff
@@ -53,10 +53,29 @@ type InfluxDB struct {
...
+	case reflect.Uint:
+	case reflect.Uint8:
+	case reflect.Uint16:
+	case reflect.Uint32:
+	case reflect.Uint64:
+	case reflect.Uintptr:
+	case reflect.Float32:
+	case reflect.Float64:
+	case reflect.String:
+	default:
```
sklirg: Wouldn't all these fall through to the default case which errors out?

Issue #82, opened 2019-11-19
Support loading configuration from multiple files


KristianLyng:
As configuration grows, it is useful to provide a set of standard module configurations that can be easily moved around.

Supporting loading of multiple configuration files should be fairly straight forward. If Bytes() is allowed to take an array of arrays of bytes, unmarshalling all of them before secondPass should Just Work™, though there are some minor issues. One of them is what to do with overlap... An alternative is just using different config's I suppose, it shouldn't be too difficult to get the core functionality up.

PR #121, opened 2020-03-02
Multi file config

sklirg:
Should be pretty self explanatory, but I'll add a quick note.

This change request adds the feature detailed in #82 which suggests adding support for reading a directory for configuration files.

This is added by a new program flag (currently -d but that can be changed as it's usually used for --debug etc), which calls into ReadFiles from the configuration package. This reads all files ending in `.json`, parses them as JSON (and will fail if it's not valid JSON). After identifying and reading the files, it will marshal it back to JSON and then unmarshal it into a larger configuration object, after which it runs the normal Bytes()-configuration parsing.

There's no need to define each section in each file, so it's possible to only define what you want (e.g. a single receiver). Inside each section, a key is unique. This is the same as in regular JSON (and regular skogul config), defining the same key multiple times is ambiguous and wouldn't work today.

### DIFF
--- a/cmd/skogul/main.go
+++ b/cmd/skogul/main.go
@@ -48,6 +48,7 @@ import (
 var versionNo string
 
 var ffile = flag.String("f", "~/.config/skogul.json", "Path to skogul config to read.")
+var fconfigDir = flag.String("d", "", "Path to skogul configuration files.")
 var fhelp = flag.Bool("help", false, "Print more help")
 var fconf = flag.Bool("show", false, "Print the parsed JSON config instead of starting")
 var fman = flag.Bool("make-man", false, "Output RST documentation suited for rst2man")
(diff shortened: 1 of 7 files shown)


sklirg:
I started stabbing at this on Friday. If I recall correctly, it's not super simple to merge the different json structures into one big map[string]interface{} (to find potential errors; second argument to secondPass). I started writing a custom unmarshaler for it, which works, but I'm basically re-implementing json.Unmarshal and adding merges for identifying if a key already exists and if so, add to it rather than replacing it.

Step 2 works though, simply unmarshalling onto an existing structure. So that's neat. But we lose all track of parsing errors if we do. Also, if we write a custom unmarshaller for creating the data passed to secondPass, we've potentially changed the structure from unmarshal into config.Config and the custom-parsed JSON.

I think the approach which makes it the easiest to debug (from a user perspective) is to either adapt secondPass/config.Config unmarshaller to have a greater understanding of what it's parsing, e.g. file name, so that can be reported back (i.e. using step2 but adding some to it), or doing it as suggested here, which merges all the JSON first and then tries to use that as a config.

I'm not sure what's best really. Without doing any big changes we would need to manually parse the json to map[string]interface{} to avoid conflicting keys anyways if we want to give some feedback to the user.

KristianLyng:
I'm closing this since #124 superseeds it.

PR #122, opened 2020-04-02
Feat/influxdb

sklirg:
Adds a simple parser for the influxdb line protocol. Should support the full influxdb line protocol, including optional timestamps as well as escaped characters in measurement names, tag/field keys and values.

Benchmarks to compare to existing parsers:

```
pkg: github.com/telenornms/skogul/parser
BenchmarkInfluxDBLineParse-16                    	 1390402	      8674 ns/op
BenchmarkInfluxDBLineParseWithoutTimestamp-16    	 1438144	      8400 ns/op
BenchmarkJSONParse-16                            	 2627970	      4612 ns/op
BenchmarkRawJSONParse-16                         	 3644203	      3286 ns/op
BenchmarkProtoBufParse-16                        	  610814	     20194 ns/op
```


sklirg:
Merging this as the code changes don't impact the current code unless the user opts into using the new influx parser. It'll be marked as unstable/beta in the release notes.

PR #123, opened 2020-04-03
feat: Support sending integers as influx-integers to influx

sklirg:


sender/influxdb.go
```diff
@@ -146,21 +147,14 @@ func (idb *InfluxDB) Send(c *skogul.Container) error {
...
-			fmt.Fprintf(&buffer, ",%s=%v", idb.replacer.Replace(key), field)
+			fmt.Fprintf(&buffer, ",%s=%v", idb.replacer.Replace(key), idb.writeValue(value))
 			nmdata++
 		}
 		fmt.Fprintf(&buffer, " ")
 		comma := ""
 		for key, value := range m.Data {
 
-			fmt.Fprintf(&buffer, "%s%s=%#v", comma, idb.replacer.Replace(key), value)
+			fmt.Fprintf(&buffer, "%s%s=%#v", comma, idb.replacer.Replace(key), idb.writeValue(value))
```
KristianLyng, line 157: This actually breaks string handling.

InfluxDB differentiates between how tag values and field values should be treated, so only tag values need to use idb.replacer.Replace() on the value, while field values are handled by double quoting, which is why one format is %v and the other %#v (I wonder what they were smoking...).

See https://docs.influxdata.com/influxdb/v1.7/write_protocols/line_protocol_reference/#quoting

sender/influxdb.go
```diff
@@ -48,6 +48,7 @@ type InfluxDB struct {
 	Measurement             string          `doc:"Measurement name to write to."`
 	MeasurementFromMetadata string          `doc:"Metadata key to read the measurement from. Either this or 'measurement' must be set. If both are present, 'measurement' will be used if the named metadatakey is not found."`
 	Timeout                 skogul.Duration `doc:"HTTP timeout"`
+	ConvertInts             bool            `doc:"Convert integers to influx line protocol integers"`
```
KristianLyng, line 51: As we've discussed, this is adding a level of technical debt - pending decisions on our end, a few thoughts:

1. The boolean should be flipped: The default behavior should be that integers get sent as integers.
2. The documentation really really REALLY should have a huge "don't do this if you don't know EXACTLY why"-warning.
3. We may want to add a skogul.Deprecated()-type function, though it's a bit premature: When this is used, we should print a warning,both because it's a generally Bad Idea in 98% of all circumstances, and because we want to remove the toggle in the future.

PR #124, opened 2020-04-27
feat: Add loading config from directory of config files

### LINKED ISSUE #82: Support loading configuration from multiple files

sklirg:
Supersedes and closes #121.

This PR implements the same as #121 just in a different way. The same startup parameter exists, `-d`, which points to a directory of configuration files to be read. Skogul will read all json files in that directory and generate a configuration from this.

Superfluous config parameters have been moved out of secondPass and into Unmarshal for the specific types/interfaces.

Closes #82.

### DIFF
--- a/config/parse.go
+++ b/config/parse.go
@@ -296,8 +296,7 @@ func findConfigFiles(path string) ([]string, error) {
 	confLog.WithField("path", path).Debugf("Reading configuration files from %s", path)
 	configFiles := make([]string, 0)
 	err := filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
-		if !info.IsDir() && len(strings.Split(path, ".")) > 1 &&
-			strings.ToLower(strings.Split(path, ".")[1]) == "json" {
+		if !info.IsDir() && filepath.Ext(path) == ".json" {
 			configFiles = append(configFiles, path)
 		}
 		return err
--- a/config/parse.go
+++ b/config/parse.go
@@ -30,6 +30,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"path/filepath"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


config/parse.go
```diff
@@ -276,6 +292,57 @@ func File(f string) (*Config, error) {
 	return Bytes(dat)
 }
 
+func findConfigFiles(path string) ([]string, error) {
+	confLog.WithField("path", path).Debugf("Reading configuration files from %s", path)
+	configFiles := make([]string, 0)
+	err := filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
+		if !info.IsDir() && len(strings.Split(path, ".")) > 1 &&
+			strings.ToLower(strings.Split(path, ".")[1]) == "json" {
```
KristianLyng: This will fail if the file is name "foo.bar.zoo.json" unless I'm reading this wrong. 

Can't we just check that len(strings) > len(".json") then check if strings[len(strings)-len(".json"):] == ".json" or something like that?

KristianLyng: Hmm, also, walk doesn't follow symbolic links? That's.... problematic and downright stupid, but we can live with that _for now_.

KristianLyng (changes requested):
You've got a thumbs up on the important bits from me here, this looks good. (Sorry for the delayed review).

The only thing I am requesting a small change with regards to is how we look for ".json" in the filenames as I suspect it will break with filenames with multiple dots in them.

config/parse.go
sklirg: I've resolved the original comment, but walk still works as walk does. We might look into that in the future as you say.

PR #125, opened 2020-07-01
influx-parser: Allow '=' in tag value

sklirg:
I found a bug which would abort the parsing process of tags, resulting in the metric containing no useful tags (other than the measurement tag which is "manually" added to the list).

The code splits each tag on '=' after having parsed it into key/value pairs, so when it split on '=' it naively expected there to only be two values in the resulting list, and if not, it would stop parsing altogether - possibly expecting this to be the end of the list of tags, or trying to safely recuperate after an error. If the tag value contains more '='s, we should accept that and put it back into the tag value.

Now we split on the **first** '=' and then add the rest of the string back as the tag value.

Note: This does not add support for '=' in the tag key. I have yet to see this, so I'll leave that for later and create an issue for it.


sklirg:
#126 for tracking the note.

parser/influxdb.go
```diff
@@ -142,11 +142,11 @@ func (line *InfluxDBLineProtocol) ParseLine(s string) error {
 		tag := strings.Trim(tagScanner.Text(), "\u0000")
 		tagValue := strings.Split(tag, "=")
 
-		if len(tagValue) != 2 {
+		if len(tagValue) < 2 {
 			break
 		}
 
-		line.tags[tagValue[0]] = tagValue[1]
+		line.tags[tagValue[0]] = strings.Join(tagValue[1:], "=")
```
KristianLyng: This can really be done with strings.SplitN(tag,"=",2), which will return two substrings, where ` the last substring will be the unsplit remainder.`.

This is considerably faster, since all processing will stop at the first =, and no join is needed. Whether we will notice the difference or not is a different matter.

KristianLyng (changes requested):
Split+Join seems redundant when we can do SplitN instead to achieve the same result. Otherwise, feel free to merge. Even with the existing implementation I'm ok with it if you're feeling lazy.

parser/influxdb.go
sklirg: I'll check out SplitN.

Join will only be a performance impact if the remaining number of strings is > 1. Otherwise Join will simply return the first string in the list. Sure, it'll still do a comparison for list length, but I would've done that in our code myself. No comparisons sounds like would be even better, though, and string-building might be slow.

PR #130, opened 2020-09-03
feat: Implement file sender

sklirg:
Adds a file sender. The configuration options support truncating the file on startup, or leaving it as it is, as well as specifying a path to the file.

I have yet to write tests but I am a bit unsure of where to start in that regard. I'll scour the other senders and their tests to see if I can come up with something.

### DIFF
--- a/sender/file.go
+++ b/sender/file.go
@@ -72,6 +72,7 @@ func (f *File) init() {
 	// Listening to a channel is blocking so we have
 	// to start the channel listening in a goroutine
 	// so that init() doesn't block.
+	f.c = make(chan []byte)
 	go f.startChan()
 
 	f.ok = true
@@ -81,7 +82,6 @@ func (f *File) startChan() error {
 	fileLog.Trace("Starting file writer channel")
 	// Making sure we close the file if this function exits
 	defer f.f.Close()
-	f.c = make(chan []byte)
 	for b := range f.c {
 		written, err := f.f.Write(append(b, newLineChar))
 		if err != nil {
--- a/sender/file.go
+++ b/sender/file.go
@@ -72,7 +72,7 @@ func (f *File) init() {
 	// Listening to a channel is blocking so we have
 	// to start the channel listening in a goroutine
 	// so that init() doesn't block.
-	f.c = make(chan []byte)
+	f.c = make(chan []byte, 50)
 	go f.startChan()
 
 	f.ok = true
--- a/sender/file.go
+++ b/sender/file.go
@@ -1,0 +1,126 @@
+/*
+ * skogul, file writer
+ *
+ * Copyright (c) 2019 Telenor Norge AS
+ * Author(s):
+ *  - Håkon Solbjørg <Hakon.Solbjorg@telenor.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


sklirg:
Should we have support for different output formats? Currently, every skogul container is written as json on a single line in the file. Should we support pretty printing the json? Should we support some form of csv/tsv, or something else we currently can parse (influx, sql)?

If so, I'd vouch for adding this as a separate issue.

sender/file.go
```diff
@@ -0,0 +1,126 @@
...
+		fileLog.Trace("Creating file since it doesn't exist or we don't want to append to it")
+		file, err = os.Create(f.Path)
+	}
+	if err != nil {
+		fileLog.WithError(err).Errorf("Failed to open '%s'", f.Path)
+		f.ok = false
+		return
+	}
+
+	// startChan() handles closing the file as this function returns
```
sklirg, line 68: I'm not sure if this is "good enough", or if `init()` should `Close()` the file and let `startChan()` handle (re)opening it. This makes the `defer file.Close()` look more neat and probably handles errors a bit better in the case of something going haywire during init (between `f.f = file` and when `defer f.f.Close()` happens in `startChan()`).

KristianLyng: I don't see any logical reason why we should close and re-open the file, that just creates a mess, and there's a risk that the second open() fails for any number of reasons. IMO the current implementation is solid.

sender/file.go
```diff
@@ -0,0 +1,126 @@
...
+	go f.startChan()
+
+	f.ok = true
+}
+
+func (f *File) startChan() error {
+	fileLog.Trace("Starting file writer channel")
+	// Making sure we close the file if this function exits
+	defer f.f.Close()
+	f.c = make(chan []byte)
```
KristianLyng: This needs to happen before the go routine is set up, so in init(). This is currently a race condition as the channel might not have been created yet when we try to send to it, depending on scheduling. I'd also argue it would make sense to add a size here to make it buffered, since the write-end of the channel doesn't evaluate any return value anyway. (Basically any size would work, just to prevent this from blocking. So 5/10/50 ... etc)

sender/file.go
```diff
@@ -0,0 +1,126 @@
...
+	for b := range f.c {
+		written, err := f.f.Write(append(b, newLineChar))
+		if err != nil {
+			f.ok = false
+			fileLog.WithError(err).Errorf("Failed to write to file. Wrote %d of %d bytes", written, len(b))
+			return err
```
KristianLyng: Hmm, ok, f.ok is set to false so we don't keep trying, but the error is ignored though, and no good recovery...

Ok, I figure the "correct" error-handling here is more hassle than it's worth (e.g.: periodic retries and whatnot), so this is probably fine. But IMHO we should not return an error when the error is discarded anyway, but that's a minor nit-pick.

sender/file.go
```diff
@@ -0,0 +1,126 @@
...
+
+	f.c <- b
+
+	return nil
+}
+
+// Verify checks that the configuration options are set appropriately
+func (f *File) Verify() error {
+	fileLog.Debug("Verified file sender")
+	return nil
```
KristianLyng, line 127: Since we don't _actually_ verify anything, it's probably better to simply not implement the interface (it's optional, per config/parse.go:472). But I suppose we might want to check if it would be possible to open the file and that the file-name is non-empty, but that's not really a priority IMO.

KristianLyng (changes requested):
Looks good to me, as long as you bump the channel-creation up to init(). The rest is mostly nitpicks that I consider optional to address.

sender/file.go
sklirg: I think it makes more sense to just not return anything from this non-exposed function as it's only used by a goroutine, which will disregard any errors it gets back as you mention. I agree that the error handling could be better, but at the same time I wouldn't really want errors in a single sender panic the whole runtime and we don't have any better ways of notifying the owner currently.

PR #131, opened 2020-09-09
Splunk sender

sklirg:
Still awaiting a response regarding the batching, but I thought I could get some feedback on the general look on this as it brings some changes to other parts of skogul anyways.

### DIFF
--- a/sender/http.go
+++ b/sender/http.go
@@ -121,17 +121,20 @@ func (ht *HTTP) init() {
 	if ht.Headers == nil {
 		ht.Headers = make(map[string]string)
 	}
-	hasContentTypeHeader := false
-	for header, _ := range ht.Headers {
+	contentTypeHeaderKey := ""
+	contentTypeHeaderVal := "application/json"
+	for header, val := range ht.Headers {
 		if http.CanonicalHeaderKey(header) == http.CanonicalHeaderKey("content-type") {
-			hasContentTypeHeader = true
+			contentTypeHeaderKey = header
+			contentTypeHeaderVal = val
 			break
 		}
 	}
-	// Use application/json as content-type by default.
-	if !hasContentTypeHeader {
-		ht.Headers[http.CanonicalHeaderKey("content-type")] = "application/json"
+	if contentTypeHeaderKey != http.CanonicalHeaderKey(contentTypeHeaderKey) {
+		// Enforce setting the header in the canonicalized way
+		delete(ht.Headers, contentTypeHeaderKey)
 	}
+	ht.Headers[http.CanonicalHeaderKey("content-type")] = contentTypeHeaderVal
 
 	tran := http.Transport{
 		TLSClientConfig: &tls.Config{
--- a/sender/http.go
+++ b/sender/http.go
@@ -127,22 +127,16 @@ func (ht *HTTP) init() {
 	ht.ok = true
 }
 
-// Send POSTS data
-func (ht *HTTP) Send(c *skogul.Container) error {
-	ht.once.Do(func() {
-		ht.init()
-	})
+// SendBytes uses a configured HTTP client to
+// send a request.
+// This makes it possible for other senders to
+// reuse the HTTP sender options without having
+// to re-implement them.
+func (ht *HTTP) SendBytes(b []byte) error {
 	if !ht.ok {
-		e := skogul.Error{Source: "http sender", Reason: "initialization failed"}
-		httpLog.Print(e)
-		return e
+		return skogul.Error{Reason: "HTTP sender not in OK state", Source: "http-sender"}
 	}
-	b, err := json.Marshal(*c)
-	if err != nil {
-		e := skogul.Error{Source: "http sender", Reason: "unable to marshal JSON", Next: err}
-		httpLog.WithError(e).Error("Configuring HTTP sender failed")
-		return e
-	}
+
 	var buffer bytes.Buffer
 	buffer.Write(b)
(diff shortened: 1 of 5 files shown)


sender/splunk.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/telenornms/skogul"
+)
+
+var splunkLog = skogul.Logger("sender", "splunk")
+
+// Splunk contains the configuration parameters for this sender.
+type Splunk struct {
+	URL           string `doc:"URL to Splunk HTTP Event Collector (HEC)"`
+	Token         string `doc:"Token for HTTP Authorization header for HEC endpoint."`
+	Batch         bool   `doc:"Batch multiple metrics together when sending. Default: false"`
```
KristianLyng: IMO this should not be a feature toggle. It negates the batch-sender. Disabling batching can be done with the batch-sender set to batch 1 item.

BUT, I am inclined to defer making that decision until we have established how the HEC batching works.

sender/splunk.go
```diff
@@ -0,0 +1,190 @@
...
+	Index         string `doc:"Custom Splunk index to send event to."`
+	HostnameField string `doc:"Name of the metadata field with the hostname. Note, this might have to be transformed into metadata depending on the input data."`
+	HTTP          *HTTP  `doc:"HTTP sender options. URL is overwritten from this config, the rest will be HTTP sender defaults unless overridden."`
+	ok            bool
+	once          sync.Once
+}
+
+// SplunkEvent describes the structure of a Splunk
+// HTTP Event Collector event
+type SplunkEvent struct {
```
KristianLyng: Does this need to be exported? If not: lower-case....

sender/splunk.go
```diff
@@ -0,0 +1,190 @@
...
+	Host  string                 `json:"host,omitempty"`
+	Index string                 `json:"index,omitempty"`
+	Event map[string]interface{} `json:"event"`
+}
+
+// prepare converts a skogul container into the appropriate
+// format expceted by the Splunk HEC collector as defined here
+// https://docs.splunk.com/Documentation/Splunk/8.0.6/Data/FormateventsforHTTPEventCollector
+func (s *Splunk) prepare(c *skogul.Container) ([]SplunkEvent, error) {
+	events := make([]SplunkEvent, 0)
```
KristianLyng: Performance nitpick: IIRC we can create it with the correct size right out of the box. make([]SplunkEvent,0, len(c.Metrics)) should do it (length 0 so we can append normally, but capacity correct).

sender/http.go
```diff
@@ -166,7 +167,16 @@ func (ht *HTTP) Send(c *skogul.Container) error {
 		}
 		ht.client = &http.Client{Transport: &tran, Timeout: ht.Timeout.Duration}
 	})
-	resp, err := ht.client.Post(ht.URL, "application/json", &buffer)
+	req, err := http.NewRequest("POST", ht.URL, &buffer)
```
KristianLyng, line 184: I get the transition to NewRequest, but for the HTTP sender itself we need to provide the Content-Type application/json at least. It can be done in init, though: After headers are parsed, check if Content-Type is set and add application/json if it isn't set. That will give the user the ability to override the content-type too.

KristianLyng (commented):
I think the approach with nesting/reusing the HTTP sender is ... intriguing and makes a lot of sense. I need to/want to review and analyze it further before final merge, sort of "big picture", but so far I'm positive.

I made some minor comments, the biggest is the "optional" batching which I _want_ to keep out of the splunk-sender itself, but let's see how batching actually works. POLA definitely dictates that regardless of what the batch sender is set to, the data will eventually be stored with the same content, and if the batching of HEC is as weird as we fear, that isn't the case and then internal batch-enable/disable toggles make sense.... we'll see I suppose.

sender/splunk.go
```diff
@@ -0,0 +1,174 @@
...
+
+var splunkLog = skogul.Logger("sender", "splunk")
+
+// Splunk contains the configuration parameters for this sender.
+type Splunk struct {
+	URL           string `doc:"URL to Splunk HTTP Event Collector (HEC)"`
+	Token         string `doc:"Token for HTTP Authorization header for HEC endpoint."`
+	Index         string `doc:"Custom Splunk index to send event to."`
+	HostnameField string `doc:"Name of the metadata field with the hostname. Note, this might have to be transformed into metadata depending on the input data."`
+	HTTP          *HTTP  `doc:"HTTP sender options. URL is overwritten from this config, the rest will be HTTP sender defaults unless overridden."`
```
KristianLyng, line 45: The documentation should probably state that this uses the config from the HTTP sender. I'm starting to think we might need/want a richer way to do documentation for certain modules. Not for the individual options, since that would bloat it, but a better way to provide examples/prose of a generic nature. Not relevant for this PR though, just a brain dump.

KristianLyng (changes requested):
The more I think about it, the more sense this HTTP-thing makes.

I had been coming to realize that we needed a sender-equivalent of the "parser"-logic, which is exactly what this accomplishes, in a less generic manner.  But making a completely generic solution like the parsers are for receivers isn't really possible, because as Splunk demonstrates here, there are needs for extra implementation-specific tweaks to the transport, e.g.: HTTP headers.

So from what I can tell, this is a good approach.

My only tiny little regret is the dual definition of the URL. I think Verify should probably check if both are specified and give a warning if there's a missmatch, at the very least.

Other than that, I think we should merge this and get some mileage on it.

PR #133, opened 2020-09-18
Create resolveParser func to resolve parsers

sklirg:
Extracts the if/elseif/else chain for finding a parser to its own function.

### DIFF
--- a/config/parse.go
+++ b/config/parse.go
@@ -371,22 +371,12 @@ func resolveHandlers(c *Config) error {
 
 		h.Handler.Sender = h.Sender.S
 		h.Handler.Transformers = make([]skogul.Transformer, 0)
-		if h.Parser == "protobuf" {
-			logger.Debug("Using protobuf parser")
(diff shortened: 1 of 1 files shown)


KristianLyng (commented):
Hmm, this isn't too horrible, and is OK, but I think it means it's time we promote parsers to full modules. It should be a simple matter of just adding the logic similar to sender/auto.go to parser/auto.go, then using that instead of a hardcoded list. It will also give us more consistent documentation, which has been a (minor) issue for parsers. 

I also think it's time for "default" modules. E.g.: no parsers have any options, so referencing a parser by the implementation name (e.g.: json/influx) should silently instantiate a json/influx parser if it doesn't exist. This could be the same for other modules, such as the template transformer and the debug sender.

But since that's a somewhat bigger project, I'm ok with merging this PR now and adding issues for the above actions. I leave that decision to you, but the PR itself is A-OK.

PR #134, opened 2020-09-18
parser: Implement parser for MNR data

sklirg:
Depends on the changes in #133 for adding the appropriate parser, other than that - it's a pretty simple new parser.


parser/mnr.go
```diff
@@ -0,0 +1,204 @@
...
+	"unicode/utf8"
+
+	"github.com/telenornms/skogul"
+)
+
+var mnrLog = skogul.Logger("parser", "mnr")
+
+type MNR struct{}
+
+const MNRSeparator byte = 9 // tab
```
KristianLyng: Should this really be exported?

parser/mnr.go
```diff
@@ -0,0 +1,204 @@
...
+		if err != nil {
+			// If we get an error on this line we will continue
+			// in hopes of having other lines working successfully.
+			mnrLog.WithError(err).Error("Failed to parse MNR line")
+			continue
+		}
+		metrics = append(metrics, metric)
+	}
+	if len(metrics) == 0 {
+		mnrLog.WithField("lines", len(lines)).Warnf("MNR parser failed to parse all of the %d lines", len(lines))
```
KristianLyng: This is an incorrect error message. It's checking if it parsed _any_ messages, not "all".

I'm not sure what the correct response should be if we have 0 viable metrics. Returning an empty array seems inconvenient - it will cause a validation error due to the check in data.go:172 that checks if the metrics array is empty.

KristianLyng (approved):
Looks good to me. I can't really comment too much on the actual parse-logic, but it _looks_ sensible.

parser/mnr.go
sklirg: Fixed message.

The correct thing would be to error, I think, and both empty array and "nil" will do that. Nil might be more appropriate.

parser/mnr.go
sklirg: No.

Issue #136, opened 2020-09-22
Promote parsers to full modules


KristianLyng:
We now have more than a couple of parsers and it's time to promote them to regular modules.

This means adding parser/auto.go and using /modules.go-logic.

This should probably be done in conjunction with #135 since none of the parsers have any options.

Still, the core config should support having a "parsers"-section for future-compatibility and to align parsers with senders, receivers and transformers.

KristianLyng:
Fixed with #137

PR #137, opened 2020-09-23
Promote parsers to fully fledged modules

### LINKED ISSUE #135: Modules should have a boolean for default-empty instantiation
Several modules don't require any options, or will work fine without them. Specially with parsers now inching closer to being full modules, this is highly relevant.

Today, both the debug sender and the template transfromer are examples of full modules that have 0 configurable options. As such, having a list of empty modules just to instantiate them seems unnecessary.

My suggestion is we add a boolean to modules that, when set, tells the config parser to instantiate an empty version of this module for each of the available names. That way, you can refer to "debug" sender without having a ```"senders": { "debug": { "type": "debug" }}```.

Some modules that even support options might benefit from this too. E.g.: The file receiver can have a "stdin"-variant available with a custom alloc-function that set it up to read from stdin.

This should be a general-purpose module thing, not one-offs.

KristianLyng:
This ALSO introduces the concept of "automake" modules, which are
modules that do not need to be explicitly defined in their relevant
configuration section. This will let us continue to refer to parsers as
just "json" without having to explicitly define the json-parser, which
doesn't require any options, but gives us the option to do so.

In addition to most parsers, this has been enabled for the debug-sender
and the templater transformer.

The rest of this patch is mostly just a matter of replacing the
hardcoded resolveParser() with the resolveParsers() function that is
similar to resolveSenders, etc, and the required scaffolding to support
that with various Marshal/Unmarshal functions.

All tests pass.

Fixes #135 #136 

NOTE: This introduces one potentially breaking change: In the past, you
could omit the parser from the handler section and it would silently
default to json. This is no longer allowed and a parser MUST be
explicitly configured for every handler. Now that we have a decent
amount of parsers, this seems justified.

### DIFF
--- a/cmd/skogul/main.go
+++ b/cmd/skogul/main.go
@@ -1,9 +1,10 @@
 /*
  * skogul, using config file
  *
- * Copyright (c) 2019 Telenor Norge AS
+ * Copyright (c) 2019-2020 Telenor Norge AS
  * Author(s):
  *  - Kristian Lyngstøl <kly@kly.no>
+ *  - Håkon Solbjørg <hakon.solbjorg@telenor.com>
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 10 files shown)


sklirg (approved):
LGTM :shipit:

PR #141, opened 2020-10-06
Add worker-pool logic to the UDP receiver

KristianLyng:
This was harder than I had anticipated to test.

I'm exposing both the backlog size/buffering of the channel, and the
number of threads to use. To get a meaningful test, I added the
sleep-sender into the test-loop. Without it, we're testing the wrong
thing.

It turns out the backlog size isn't horribly important, which shouldn't
have surprised me to be honest. We already HAVE a backlog between in the
kernel before our call to Read. Even a small buffer would help here, but
defaulting to 100 messages felt like a reasonable value.

Number of threads is really the key factor. Since each thread will be
carried over to a sender, if we don't use a batch-sender to decouple
receiver-go routines from sender-go routines, the receiver go routine
will block waiting for the sender to complete.  E.g.: A slow sender is
actually more detrimental than a slow receiver in this case.

PR attaches some performance data from test cases, pre- and post this
fix. This fix is not really meant to be a huge performance boost, but a
means to prevent the UDP receiver from killing the rest of Skogul if the
sender pipeline is slower than the amount of messages received. In the
past, this would just spawn endless go routines, now, instead, we will
simply drop packages as the receive buffer fills up.

skogul-udp-bench

Ideally we should check for that, but we didn't to begin with so this
isn't really a regression. And I'm honestly not sure if it's possible.

### DIFF
--- a/receiver/udp.go
+++ b/receiver/udp.go
@@ -25,6 +25,7 @@ package receiver
 
 import (
 	"net"
+	"runtime"
 
 	"github.com/telenornms/skogul"
 )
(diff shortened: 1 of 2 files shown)


sklirg (approved):
LGTM :shipit:

PR #142, opened 2020-10-07
Authentication using HTTP Client Certificates

### LINKED ISSUE #40: HTTP receiver: verify client certificates
Need to be able to verify client certificates, possibly in addition to username/password. Should also support alternate root CA in such scenarios.

### LINKED ISSUE #41: HTTP sender: Use client certificates
The other side of #40. Need to be able to authenticate with client certificates.

sklirg:
Fixes #40, fixes #41

### DIFF
--- a/receiver/http.go
+++ b/receiver/http.go
@@ -42,8 +42,11 @@ var httpLog = skogul.Logger("receiver", "http")
 
 // HTTPAuth contains ways to authenticate a HTTP request, e.g. Username/Password for Basic Auth.
 type HTTPAuth struct {
-	Username string `doc:"Username for basic authentication. No authentication is required if left blank."`
-	Password string `doc:"Password for basic authentication."`
+	Username              string `doc:"Username for basic authentication. No authentication is required if left blank."`
+	Password              string `doc:"Password for basic authentication."`
+	SANDNSName            string `doc:"DNS name which has to be present in SAN extension of x509 certificate when using Client Certificate authentication"`
+	SkipCertificateVerify bool   `doc:"Skip verifying certificate. (default: false)"`
+	path                  string
 }
 
 /*
@@ -58,7 +61,6 @@ type HTTP struct {
 	Certfile             string                        `doc:"Path to certificate file for TLS. If left blank, un-encrypted HTTP is used."`
 	Keyfile              string                        `doc:"Path to key file for TLS."`
 	ClientCertificateCAs []string                      `doc:"Paths to files containing CAs which are accepted for Client Certificate authentication."`
-	SANDNSName    string                        `doc:"DNS name which has to be present in SAN extension of x509 certificate when using Client Certificate authentication"`
 }
 
 // For each path we handle, we set up a receiver such as this
@@ -84,6 +86,15 @@ func (auth *HTTPAuth) auth(r *http.Request) error {
 		}
 
 		return nil
+	}
+
+	if auth.SANDNSName != "" {
+		httpLog.Trace("Verifying request using client certificates")
+		if err := auth.verifyPeerCertificate(nil, r.TLS.VerifiedChains); err != nil {
+			return err
+		} else {
+			return nil
+		}
 	}
 
 	return skogul.Error{Source: "http receiver", Reason: "No matching authentication method"}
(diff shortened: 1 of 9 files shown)


receiver/http.go
```diff
@@ -148,6 +168,19 @@ func (htt *HTTP) Start() error {
...
 
+	if len(htt.AcceptableCAs) > 0 {
+		pool, err := loadClientCertificateCAs(htt.AcceptableCAs)
+		if err != nil {
+			httpLog.WithError(err).Error("Failed to load Client Certificates")
+			return err
+		}
+		server.TLSConfig = &tls.Config{
+			ClientCAs:  pool,
+			ClientAuth: tls.RequireAndVerifyClientCert,
```
sklirg: Should the policy of authentication be configurable? If so, which auth types should we support?

https://golang.org/pkg/crypto/tls/#ClientAuthType

receiver/http.go
```diff
@@ -133,6 +137,22 @@ func (rcvr receiver) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	rcvr.answer(w, r, code, err)
 }
 
+// loadClientCertificateCAs loads a given list of strings as paths of
+// acceptable CAs for use in Client Certificate authentication.
+func loadClientCertificateCAs(paths []string) (*x509.CertPool, error) {
+	httpLog.Debugf("Loading %d Client Certificate(s) from file(s)", len(paths))
```
KristianLyng: This is probably incorrect, isn't it? It's common to store multiple certificates in a single file (my /etc/ssl/certs/ca-certificates.crt has 127 ca certificates). So the message should probably read "Loading %d certificate files" or something like that. Perhaps somewhat nitpicky, but it can certainly create confusion during troubleshooting.

receiver/http.go
```diff
@@ -47,11 +50,12 @@ least one handler to be set up, using Handle. This is done implicitly
...
-	Auth     map[string]*HTTPAuth          `doc:"A map corresponding to Handlers; specifying authentication for the given path, if required."`
-	Certfile string                        `doc:"Path to certificate file for TLS. If left blank, un-encrypted HTTP is used."`
-	Keyfile  string                        `doc:"Path to key file for TLS."`
+	Address       string                        `doc:"Address to listen to." example:"[::1]:80 [2001:db8::1]:443"`
+	Handlers      map[string]*skogul.HandlerRef `doc:"Paths to handlers. Need at least one." example:"{\"/\": \"someHandler\" }"`
+	Auth          map[string]*HTTPAuth          `doc:"A map corresponding to Handlers; specifying authentication for the given path, if required."`
+	Certfile      string                        `doc:"Path to certificate file for TLS. If left blank, un-encrypted HTTP is used."`
+	Keyfile       string                        `doc:"Path to key file for TLS."`
+	AcceptableCAs []string                      `doc:"Paths to files containing CAs which are accepted for Client Certificate authentication."`
```
KristianLyng: I'm not sure I like this naming. And I'm curious if we can also use the default system-wide CA? In that case, we could/would need two options. One to enable client certificate validation, the other to list custom CAs... That might be overkill though. Thoughts?

As for the name, regardless of if it's one or two: What about ClientCertificateCAs ? The reason I'm skeptical to AcceptableCAs is that client certificates is somewhat obscure to most people and as such just saying "AcceptableCAs" could mislead people to think it's about server-side certificates.

receiver/http.go
KristianLyng: See previous comment, but I think "nyeso".

It makes very little sense to accept and/or require Client certificates without verifying them, so the only valid setting I can imagine is disabling it completely (the default) or RequireAndVerifyClientCert.

sender/http.go
```diff
@@ -95,13 +97,22 @@ func getCertPool(path string) (*x509.CertPool, error) {
...
+}
+
 func (ht *HTTP) init() {
 	ht.ok = false
 	if ht.Timeout.Duration == 0 {
 		ht.Timeout.Duration = 20 * time.Second
 	}
 	if ht.Insecure {
-		httpLog.Print("Warning: Disabeling certificate validation for HTTP sender - vulnerable to man-in-the-middle")
+		httpLog.Warning("Disabling certificate validation for HTTP sender - vulnerable to man-in-the-middle")
```
KristianLyng, line 116: Nice catch :D

sender/http.go
```diff
@@ -111,15 +122,29 @@ func (ht *HTTP) init() {
...
+		httpLog.WithError(err).Error("Failed to initialize root CA pool")
 		return
 	}
 
+	tlsConfig := &tls.Config{
+		InsecureSkipVerify: ht.Insecure,
+		RootCAs:            cp,
+	}
+
+	if ht.Certfile != "" && ht.Keyfile != "" {
```
KristianLyng, line 135: Since both are required if either is specified, we should probably check for this in Verify() to make sure.

KristianLyng (changes requested):
I am fairly happy with this, but I haven't actually tested it. It seems to do what it needs to do.

I am honestly not sure if having two options in the receiver instead of one (see discussion) is better, so I'll leave that decision to you. It can always be introduced later on. Come to think of it, since we're unsure, I think just having the simplest implementation (the one you suggest) is the best way to move forward.

Other than that, I would prefer a name change for AcceptableCAs, but it's not a requirement.

However, for the sender, we should Verify() that either both key/cert is provided, or neither, to avoid user-confusion. Since that's trivial, that's the only thing I'm really making a requirement from my thumbs up.

KristianLyng (approved):

PR #143, opened 2020-10-09
feat: Verify Client Certificate

### LINKED ISSUE #81: HTTP receiver: Authorize based of contents of client certificate
#40 is about authenticating with a client certificate, but this issue should handle how we can carry out authorization based on client certificates.

E.g.: Even if a client has a valid certificate, we want to be able to distinguish one client from an other.

This depends on #40 being completed.

sklirg:
Depends on #142.

Fixes #81 

Only includes verification checks of DNS names.

### DIFF
--- a/receiver/http.go
+++ b/receiver/http.go
@@ -41,8 +41,11 @@ var httpLog = skogul.Logger("receiver", "http")
 
 // HTTPAuth contains ways to authenticate a HTTP request, e.g. Username/Password for Basic Auth.
 type HTTPAuth struct {
-	Username string `doc:"Username for basic authentication. No authentication is required if left blank."`
-	Password string `doc:"Password for basic authentication."`
+	Username              string `doc:"Username for basic authentication. No authentication is required if left blank."`
+	Password              string `doc:"Password for basic authentication."`
+	SANDNSName            string `doc:"DNS name which has to be present in SAN extension of x509 certificate when using Client Certificate authentication"`
+	SkipCertificateVerify bool   `doc:"Skip verifying certificate. (default: false)"`
+	path                  string
 }
 
 /*
@@ -57,7 +60,6 @@ type HTTP struct {
 	Certfile             string                        `doc:"Path to certificate file for TLS. If left blank, un-encrypted HTTP is used."`
 	Keyfile              string                        `doc:"Path to key file for TLS."`
 	ClientCertificateCAs []string                      `doc:"Paths to files containing CAs which are accepted for Client Certificate authentication."`
-	SANDNSName    string                        `doc:"DNS name which has to be present in SAN extension of x509 certificate when using Client Certificate authentication"`
 }
 
 // For each path we handle, we set up a receiver such as this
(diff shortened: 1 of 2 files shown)


sklirg:
One interesting caveat: This verification has to happen before "HTTP", which means that we don't know what URL the user is accessing at the point of verification. In turn, this means that our current "auth"-dict which is based on `/path`s will not work for verifying using client certificates.

Potentially, we'll have to create separate http receivers for each handler. It's not a problem, really, it just doesn't work as well with our existing auth-scheme.

receiver/http.go
```diff
@@ -56,6 +57,7 @@ type HTTP struct {
 	Certfile      string                        `doc:"Path to certificate file for TLS. If left blank, un-encrypted HTTP is used."`
 	Keyfile       string                        `doc:"Path to key file for TLS."`
 	AcceptableCAs []string                      `doc:"Paths to files containing CAs which are accepted for Client Certificate authentication."`
+	SANDNSName    string                        `doc:"DNS name which has to be present in SAN extension of x509 certificate when using Client Certificate authentication"`
```
sklirg: We could try to work this more into the style of HTTPAuth (line 42-46), as adding struct fields for each of the potential ways to verify (including DNSNames, EmailAddresses, IPAddressess, URIs) could become a bit messy if all added onto the base HTTP struct.

Just a thought.

receiver/http.go
```diff
@@ -192,6 +198,43 @@ func (htt *HTTP) Start() error {
 	return skogul.Error{Reason: "Shouldn't reach this"}
 }
 
+// verifyPeerCertificate verifies a client certificate presented to us
+// during TLS handshake by comparing its extensions (such as SAN) to
+// some expected value(s)
+func (htt *HTTP) verifyPeerCertificate(_ [][]byte, verifiedChains [][]*x509.Certificate) error {
+	certLogger := httpLog
```
sklirg, line 223: Logs in here should maybe be left out of the common path, i.e. opt-in through config? This could be a high throughput function which shouldn't incur any extra overhead unless requested.

receiver/http.go
```diff
@@ -192,6 +198,43 @@ func (htt *HTTP) Start() error {
...
+	certLogger.Tracef("Verifying %d certificate chains", len(verifiedChains))
+	for _, chain := range verifiedChains {
+		if len(chain) < 1 {
+			// Chain has no certificates (?)
+			continue
+		}
+		cert := chain[0]
+		certLogger := certLogger.WithFields(log.Fields{
```
KristianLyng: I am not a big fan of overshadowing variable names O_o. The logger isn't a big issue, but imagine trying to debug this at 3AM and not understanding why certLogger means something different on line 206 and line 222.

KristianLyng (changes requested):
Right, so the big thing about this is obviously the lack of granularity.

AFAICT, there's not cryptographic reason why we can't check the SAN-DNS connection after we've got the request. We should then have validated the certificate and established a TLS session both ways. 

However, I realize that this implementation is by far the most trivial since it basically offloads the entire job to the standard library.

However, did you investigate if we can authorize in ServeHTTP instead? I see that we have http.Request there, which in turn has tls.ConnectionState which IN turn has the same verifiedChains array that verifyPeerCertificate relies on. Perhaps we can simply call it manually in ServeHTTP with relevant path(s) instead of relying on it being called implicitly? At first glance, that doesn't seem like too much effort, and would allow us to have path-specific authorization again.

As for how to configure that, I'm thinking just adding SANDNSName to HTTPAuth is an alternative, though it requires some more Validate() work to be fool proof.

sklirg:
I moved the verification logic into the auth handler, as is done for basic auth. It seems to play nice.

KristianLyng (approved):
This looks good to me now.

PR #144, opened 2020-10-12
Chore: Update a ton of copyright-headers

KristianLyng:
I tried to verify that they have actually been updated in 2020, but I
figure it's no disaster if we miss. I also added Håkon to files where
git-log told me he had more than a line or two of edits, but that might
not be fool proof.


KristianLyng:
The review-request is mostly just in case I've added you to some files you know you haven't done anything with (personally I don't really mind, just figured it be rude not to let you review it)

sklirg (approved):

PR #146, opened 2020-10-12
feat: Add parse transformer

sklirg:
The parse transformer is a transformer which supports specifying a
parser to use for a specific source field, parsing that field using a
skogul parser.

~~Note: This does not support using a parser defined in "parsers". I'm thinking this should be supported, but could perhaps be left out for this first release?~~

### DIFF
--- a/transformer/auto.go
+++ b/transformer/auto.go
@@ -50,6 +50,12 @@ func init() {
 		Help:    "Enforces custom-rules for data fields of metrics.",
 	})
 	Auto.Add(skogul.Module{
+		Name:    "parse",
+		Aliases: []string{},
+		Alloc:   func() interface{} { return &Parse{} },
+		Help:    "Parses a metric using a skogul parser",
+	})
+	Auto.Add(skogul.Module{
 		Name:    "split",
 		Aliases: []string{},
 		Alloc:   func() interface{} { return &Split{} },
--- a/transformer/parse.go
+++ b/transformer/parse.go
@@ -2,60 +2,44 @@ package transformer
 
 import (
 	"fmt"
-	"strings"
 	"sync"
 
 	"github.com/telenornms/skogul"
-	"github.com/telenornms/skogul/parser"
 )
 
 var parseLog = skogul.Logger("transformer", "parser")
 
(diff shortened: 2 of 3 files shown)


transformer/parse.go
```diff
@@ -0,0 +1,123 @@
...
+	"sync"
+
+	"github.com/telenornms/skogul"
+	"github.com/telenornms/skogul/parser"
+)
+
+var parseLog = skogul.Logger("transformer", "parser")
+
+type Parse struct {
+	Parser              string `doc:"Name of skogul parser to use. Note: has to be auto-initialisable."` // Feature request: Re-use parser unmarshal logic here
```
KristianLyng: This should be a ParserRef instead of a string. That will make the config engine resolve the parser for you making that part of the PR redundant. With the implicitly created modules, it would still work just fine without requiring the user to specify a json-parser, but it will allow them to use parsers that have configuration in the future.

The only change should be changing this from string to ParserRef, removing the lower-case "parser", removing the lookup-code, and using Parser.P instead of lowercase "parser".

transformer/auto.go
```diff
@@ -49,6 +49,12 @@ func init() {
 		Alloc:   func() interface{} { return &Data{} },
 		Help:    "Enforces custom-rules for data fields of metrics.",
 	})
+	Auto.Add(skogul.Module{
+		Name:    "parse",
+		Aliases: []string{},
+		Alloc:   func() interface{} { return &Parse{} },
+		Help:    "Parses a metric using a skogul parser",
```
KristianLyng, line 56: I would perhaps add something about it being somewhat recursive. Something like "Re-parse a field in a metric using a Skogul parser. E.g.: If a metric is json-encoded, but a field in the metric is a text-string of key=value pairs, the handler will parse the JSON, and this parse transformer can be used to parse the key=value pairs." ?

KristianLyng (changes requested):
The idea behind this transformer is brilliant in its simplicity! I never even thought about it myself. 

Two things would be nice, though:

1. It needs to use ParserRef instead of doing the lookup itself.
2. There should be at least one or two test-cases, even if they are simple.

PR #147, opened 2020-10-14
feat: Add a target for releasing skogul

sklirg:
Mainly inspired by the fact that we probably should use `-a` when
tagging releases, but this way we'll have a common way of doing releases
locally. Maybe we could add a default release message too, such as
`release version v{version}` and/or the relevant section of docs/NEWS.


sklirg:
Also, this is definitely not just because we aimed for 0 open PRs by Friday.. 👀

PR #151, opened 2020-10-16
Introduce golint and fix various minor nitbits

KristianLyng:
Most of these fixes are at best semi-valid, but some inefficient
assignment fixes represent missing error-handling, so that's a nice
catch.

Making this a PR so I can verify that the auto-tests actually work with
golint.


KristianLyng:
I just added golint in drone.yml, but am I missing some obvious and more universal approach here ?

sklirg:
> I just added golint in drone.yml, but am I missing some obvious and more universal approach here ?

IIRC, `go get` installs binaries (too), so I'm sure this works fine.

One thing might be to add it to go.mod so we always get the same version, so if there are any breaking changes, our tests don't suddenly start failing.

sklirg (approved):

Issue #148, opened 2020-10-15
Make UDP packet size configurable


sklirg:
Hard-coding the UDP packet size to 9000 is not necessary. Interfaces might have bigger MTUs than that, for example loopback, which sports a massive 65k in linux.

https://github.com/telenornms/skogul/blob/v0.10.0/receiver/udp.go#L85

moogacs:
I think I am gonna work on that one, it it's ok with you

KristianLyng:
fixed by #155

PR #155, opened 2020-10-21
upd: make UDP packet size configurable #148

moogacs:
making udp packet size. configurable with taking care the UDP_MAX_READ_SIZE which is "65535" and updating the test files

### DIFF
--- a/receiver/udp.go
+++ b/receiver/udp.go
@@ -29,6 +29,7 @@ import (
 
 	"errors"
 	"github.com/telenornms/skogul"
+	"strconv"
 )
 
 const (
@@ -63,7 +64,7 @@ func (ud *UDP) process() {
 // never returns.
 func (ud *UDP) Start() error {
 	if ud.PacketSize < 0 || ud.PacketSize > UDP_MAX_READ_SIZE {
-		return errors.New("invalid udp packet size, maximum udp read size is between 0 and " + string(UDP_MAX_READ_SIZE))
+		return errors.New("invalid udp packet size, maximum udp read size is between 0 and " + strconv.Itoa(UDP_MAX_READ_SIZE))
 	}
 	if ud.Backlog == 0 {
 		ud.Backlog = 100
--- a/receiver/udp.go
+++ b/receiver/udp.go
@@ -27,7 +27,6 @@ import (
 	"net"
 	"runtime"
 
-	"errors"
 	"github.com/telenornms/skogul"
 	"strconv"
 )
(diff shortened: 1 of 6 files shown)


receiver/udp.go
```diff
@@ -56,6 +63,9 @@ func (ud *UDP) process() {
 // listening for incoming UDP messages on the configured address. Start
 // never returns.
 func (ud *UDP) Start() error {
+	if ud.PacketSize < 0 || ud.PacketSize > UDP_MAX_READ_SIZE {
+		return errors.New("invalid udp packet size, maximum udp read size is between 0 and " + strconv.Itoa(UDP_MAX_READ_SIZE))
+	}
```
sklirg: This check could be done as part of `Verify()`. I see that this receiver doesn't have it yet, but if you add it there it makes more sense, and it'll stop the program from starting up in the first place instead of erroring upon input.

`func (ud *UDP) Verify() error {}` and then return an error from there if the configuration is wrong. (Look at receivers/http.go for an example of the `Verify()` func)

You should also return the `skogul.Error` struct instead of using `errors`, putting the error message in `Reason` and write `udp-receiver` in `Source`.

receiver/udp.go
```diff
@@ -56,6 +63,9 @@ func (ud *UDP) process() {
 // listening for incoming UDP messages on the configured address. Start
 // never returns.
 func (ud *UDP) Start() error {
+	if ud.PacketSize < 0 || ud.PacketSize > UDP_MAX_READ_SIZE {
```
sklirg: Read my note above/below (regarding Verify()), and after that is fixed - I think a PacketSize  of 0 should mean "use default", and the default can probably be 9000.
This also has the benefit of test cases not needing to be touched as much, and it keeps backward compatibility.

receiver/udp.go
```diff
@@ -27,18 +27,25 @@ import (
...
-	ch      chan []byte       // Used to pass messages from the accept/read-loop to the worker pool/threads.
+	Address    string            `doc:"Address and port to listen to." example:"[::1]:3306"`
+	Handler    skogul.HandlerRef `doc:"Handler used to parse, transform and send data."`
+	Backlog    int               `doc:"Number of queued messages that are not delivered before the receiver starts blocking. Defaults to 100. Even when this is full, the kernel will still buffer data on top of this value. Higher values give smoother performance, but slightly more memory usage. Actual memory usage is a factor of average message size * backlog-fill."`
+	Threads    int               `doc:"Number of worker go routines to use, which loosely translates to parallel execution. Defaults to number of CPU threads, with a minimum of 20. There is no correct number, but the value depends on how fast your senders are."`
+	ch         chan []byte       // Used to pass messages from the accept/read-loop to the worker pool/threads.
+	PacketSize int               `doc:"UDP Packet size note: max. UDP read size is 65535"`
```
sklirg, line 46: Might be a nitpick, but I'd prefer if the private arguments are last, so if you swap the place of this and `ch`, that would be great. :)

sklirg (changes requested):
Hi there! Thanks a lot for you contribution 🎉 

I'd be glad to merge these changes if you could fix the things I commented on! :shipit: 

Do not hesitate to ask if anything is unclear. :D

receiver/udp.go
moogacs: so all comments has to fixed with last commit

sklirg (approved):
LGTM :shipit:

KristianLyng (approved):
Looks good to me, nice to get this fixed.

Issue #132, opened 2020-09-09
Custom-json parser doesn't handle json arrays


sklirg:
The custom-json parser only handles json objects, not json arrays. This is most likely due to the strict `map[string]interface{}`, which requires key/value-pairs, and does not support arrays.

If the incoming json to `custom-json` parser is on the form of 
```json
[
  {
    "foo": "bar"
  }
]
```
the custom-json parser will error with "could not parse".

I think arrays should be supported.

moogacs:
By you `map[string]interface{}` i assume you mean what is inside [parser/json.go](https://github.com/telenornms/skogul/blob/1d8fa60584c5acb117c239e4088ec99c50f013d2/parser/json.go#L55) with test transformer/[timestamp_test.go](https://github.com/telenornms/skogul/blob/1d8fa60584c5acb117c239e4088ec99c50f013d2/transformer/timestamp_test.go#L47)

KristianLyng:
Almost. The make() there is for the metadata, which is just an empty map[string]interface{} prior to any transformer filling in data. It's created in the parser because having nil-value metadata is a bit of a headache.

The real issue is https://github.com/telenornms/skogul/blob/5fbb2d6c18666420d3a60d6685b87279e98a0a81/parser/json.go#L59 where the unmarshal onto Data happens, since Data is a map[string]interface{} (Unmarshal will create it when it has a nil-value).

We don't want to change the Container defintion, so changing Data to be `interface{}` instead of `map[string]interface{}` isn't really an option.

What we could do in this case is a two-phase attempt. If the initial Unmarshal fail because the data is an array, we could create a `var foo interface{}`, Unmarshal onto that, then set up `Data["blob"] = foo`.

It would give an extra level of nesting, but I think that's reasonable in these situations. This would also mean that single-value data could be parsed, like just a single number, hence "blob" instead of "array".

My only real reservation is the extra attempt required... But I don't see any other obvious way.

KristianLyng:
This is fixed by #156

PR #156, opened 2020-10-22
fix: Custom-json parser handles json arrays #132

moogacs:
it has to be as requested and added a test case!

### DIFF
--- a/parser/json.go
+++ b/parser/json.go
@@ -59,7 +59,16 @@ func (data RawJSON) Parse(b []byte) (*skogul.Container, error) {
 	err := json.Unmarshal(b, &metric.Data)
 
 	if err != nil {
-		return nil, err
+		// Try to marshal data in an array form
+		var array []interface{}
+		err = json.Unmarshal(b, &array)
+
+		if err != nil {
+			return nil, err
+		}
+
+		metric.Data = make(map[string]interface{})
+		metric.Data["blob"] = array
 	}
 
 	container := skogul.Container{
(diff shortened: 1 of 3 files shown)


sklirg (commented):
Hi again! Thanks for another contribution :tada:

This looks to be along the lines of what was discussed in #132, which is great.

I think it could make sense to add a configuration property to the RawJSON struct, such as `IsArray: bool` if the incoming data is expected to be an array. Furthermore, this could transform the incoming array to multiple metrics, instead of putting it into the "blob" key of the first metric.

Currently, RawJSON returns a skogul.Container with 1 metric, but if we receive an array of metrics, we should return an array of metrics as part of the skogul.Container, and not only 1.

With that said, this does solve the issue outlined in #132, and to use it, users can add a transformer to split the metric into multiple metrics based on the blob field.

I think we're fine to merge this now, but I'll let @KristianLyng make the final decision. :)

KristianLyng:
I've been thinking about this.

I think I actually prefer a transformer to split the array up. I can see use-cases where data arrives as an array, but each element is not independent of the other. E.g.: You receive an array of ints that are 1-minute, 5-minute and 15-minute load averages. 

I'm merging this, because it's exactly what we discussed. A nice add-on to this would be benchmarks though. We know that unmarshal can be expensive, and I don't know how fast it can fail if it's an array. E.g.: If everything is an array, every single unmarshal would first fail, then be retried, which is obviously suboptimal, but I sort of prefer suboptimal performance to completely bloated configuration (e.g.: "ArrayFirst: true").

Thanks for the PR, good work !

PR #174, opened 2021-07-22
feat: add numThreads to batch

### LINKED ISSUE #172: Number of threads for batch sender should be configurable
It is currently using runtime.numCPU(), which should be the default, but it should also be configurable.

https://github.com/telenornms/skogul/blob/v0.10.7/sender/batch.go#L92

burntcarrot:
Resolves #172.

Changes:
- Adds numThreads to Batch struct
- Created a fillUninitialized() function to initialize fields which have not been directly filled by the user, but are needed for skogul to work properly.

Signed-off-by: burntcarrot <aadhav.n1@gmail.com>

### DIFF
--- a/sender/batch.go
+++ b/sender/batch.go
@@ -63,6 +63,7 @@ type Batch struct {
 	Next      skogul.SenderRef `doc:"Sender that will receive batched metrics"`
 	Interval  skogul.Duration  `doc:"Flush the bucket after this duration regardless of how full it is"`
 	Threshold int              `doc:"Flush the bucket after reaching this amount of metrics"`
+	Threads   int              `doc:"Number of threads for batch sender. Defaults to number of CPU cores."`
 	allocSize int
 	ch        chan *skogul.Container
 	once      sync.Once
@@ -70,7 +71,6 @@ type Batch struct {
 	timer     *time.Timer
 	cont      *skogul.Container
 	out       chan *skogul.Container
-	Threads   int `doc:"Number of threads for batch sender"`
 }
 
 func (bat *Batch) setup() {
--- a/sender/batch.go
+++ b/sender/batch.go
@@ -60,19 +60,29 @@ This means that:
 3. Send() will only block if two channels are full.
 */
 type Batch struct {
-	Next      skogul.SenderRef `doc:"Sender that will receive batched metrics"`
-	Interval  skogul.Duration  `doc:"Flush the bucket after this duration regardless of how full it is"`
-	Threshold int              `doc:"Flush the bucket after reaching this amount of metrics"`
-	allocSize int
-	ch        chan *skogul.Container
-	once      sync.Once
-	metrics   int
-	timer     *time.Timer
-	cont      *skogul.Container
-	out       chan *skogul.Container
+	Next       skogul.SenderRef `doc:"Sender that will receive batched metrics"`
+	Interval   skogul.Duration  `doc:"Flush the bucket after this duration regardless of how full it is"`
+	Threshold  int              `doc:"Flush the bucket after reaching this amount of metrics"`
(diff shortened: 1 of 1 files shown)


sender/batch.go
```diff
@@ -60,19 +60,29 @@ This means that:
...
+	Interval   skogul.Duration  `doc:"Flush the bucket after this duration regardless of how full it is"`
+	Threshold  int              `doc:"Flush the bucket after reaching this amount of metrics"`
+	allocSize  int
+	ch         chan *skogul.Container
+	once       sync.Once
+	metrics    int
+	timer      *time.Timer
+	cont       *skogul.Container
+	out        chan *skogul.Container
+	numThreads int
```
sklirg: This can be public/exported, having a capital letter. The `doc` field should also be filled in, as for "Next", "Interval" and "Threshold", so documentation is created for it.

sender/batch.go
```diff
@@ -60,19 +60,29 @@ This means that:
...
+func (bat *Batch) fillUninitialized() {
+	if bat.numThreads == 0 {
+		bat.numThreads = runtime.NumCPU()
+	}
 }
 
 func (bat *Batch) setup() {
+
+	bat.fillUninitialized()
+
```
sklirg: I think this should be done inline in the setup func instead of splitting it out into its own function.

sklirg (changes requested):
Sorry for the extremely late review! This completely blanked my mind during my holiday. Apologies!

sender/batch.go
burntcarrot: I guess you're suggesting this change:

```
NumThreads int `doc:"Number of threads for batch sender"`
```

Is this what you want it to be, or do you need any other modifications?

sender/batch.go
burntcarrot: Sounds fair, I would have incorporated this earlier, but I wasn't sure about your preferences. I'll fix this :)

burntcarrot:
> 
> 
> Sorry for the extremely late review! This completely blanked my mind during my holiday. Apologies!

Haha, no worries! Holidays are meant for having fun! ☺

sender/batch.go
sklirg: This looks good! To be in line with other configuration options regarding threads, we should name it "Threads" instead of "NumThreads", such as in the UDP receiver: https://github.com/telenornms/skogul/blob/primary/receiver/udp.go#L48

sklirg: You could also add docs about the default (default: uses the number of CPU cores). :)

burntcarrot: Ah, didn't notice this, do you want me to fix the docs? I'll add this in the next commit, if you wish.

sklirg: Yes please :)

sender/batch.go
```diff
@@ -60,29 +60,23 @@ This means that:
...
+	Interval  skogul.Duration  `doc:"Flush the bucket after this duration regardless of how full it is"`
+	Threshold int              `doc:"Flush the bucket after reaching this amount of metrics"`
+	allocSize int
+	ch        chan *skogul.Container
+	once      sync.Once
+	metrics   int
+	timer     *time.Timer
+	cont      *skogul.Container
+	out       chan *skogul.Container
+	Threads   int `doc:"Number of threads for batch sender"`
```
sklirg: You can place this above the private fields, and make sure to `gofmt` it. :)

sender/batch.go
burntcarrot: Done! 😄

sklirg (approved):
Looks good to me!

KristianLyng (approved):
Looks pretty straight forward, thanks for the pr!

PR #180, opened 2021-09-09
feat: Support for InfluxDB v2 in influx sender

### LINKED ISSUE #175: Add support for InfluxDB 2.0 in influx sender
I *think* the only requirement is to add a config field for the auth token header. The URL is generically configured so that should already be handled.

sklirg:
* Adds the configuration field "Token" to the InfluxDB sender, to support v2 of the write API.
* Changes how we create and do requests to support this change.
* Adds a "Secret"-type that can be used for configuration secrets to prevent them from being leaked.

Resolves #175.

### DIFF
--- a/common.go
+++ b/common.go
@@ -26,9 +26,10 @@ package skogul
 
 import (
 	"fmt"
-	"github.com/sirupsen/logrus"
 	"runtime"
 	"time"
+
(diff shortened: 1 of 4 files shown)


sklirg:
I'm splitting this into two new PRs.

PR #185, opened 2021-10-14
Support InfluxDB v2 in influx sender

### LINKED ISSUE #175: Add support for InfluxDB 2.0 in influx sender

sklirg:
Resolves #175.

Depends on #183.


KristianLyng (approved):
LGTM

PR #192, opened 2021-10-28
Draft: Use JSON Pointers for specifying paths in JSON documents

### LINKED ISSUE #191: Transformer: extract nested data
Ideally, we should be able to extract data according to RFC 6901, or at the very least to allow nested extraction to some degree.

E.g.: Given the structure: 
```
{
    "system":  {
           "hostname": "foo",
            "platform": "bar"
     }
}
```
We should be able to extract system/hostname with a single operation. For single-key extraction, doing something with names would be nice as well. E.g.:

```
"extractMap": {
    ["system/hostname": "hostname"],
    ["system/platform": "os"],
}
```

A more RFC 6091-ish approach overall for transformers would be nice, in fact...

sklirg:
Reworks the current "list of strings" implementation of paths in JSON documents into using the JSON Pointer package [dolmen-go/jsonptr](github.com/dolmen-go/jsonptr). Existing test cases have been updated for this change, as this is a breaking change for configuring the affected transformers.

⚠️ This will be a breaking change for the `flatten`, `timestamp` and `metadata.Flatten` transformers.

Also partly resolves #191.
Note: it does not resolve the second part of the issue mentioning "an easy way of renaming [the extracted fields]".

### DIFF
--- a/common.go
+++ b/common.go
@@ -291,21 +291,6 @@ func Assert(x bool, v ...interface{}) {
 	}
 }
 
-// ExtractNestedObject extracts an object from a nested object structure. All intermediate objects has to map[string]interface{}
-func ExtractNestedObject(object map[string]interface{}, keys []string) (map[string]interface{}, error) {
-	if len(keys) == 1 {
-		return object, nil
-	}
-
-	next, ok := object[keys[0]].(map[string]interface{})
-
-	if !ok {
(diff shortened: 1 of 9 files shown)


sklirg:
While I still believe in this implementation, I'll close it as it lags far behind primary.

PR #197, opened 2021-11-16
feat(config): Support configuring skogul with both file and directory…

### LINKED ISSUE #163: Make config-path work for both file and directory, don't need separate option for each
Let's remove the `-d` option, and repurpose `-f` into working for both config _files_ and config _directories_. We can probably figure out what the user intended by `stat`-ing the path and either load config from a file or from all files.

sklirg:
… with same flag

Resolves #163

### DIFF
--- a/cmd/skogul/main.go
+++ b/cmd/skogul/main.go
@@ -869,24 +869,17 @@ func main() {
 		os.Exit(0)
 	}
 
-	var c *config.Config
+	configPath := ""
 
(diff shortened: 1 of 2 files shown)


KristianLyng (approved):
I assume the configpath-thing still left is for backwards compatibility? Looks good to me, though.

sklirg:
Affirm. We can remove it later.

PR #198, opened 2021-11-17
Add DictSplit transformer and expand Split somewhat

KristianLyng:
The code is depressingly duplicated, save for the casting, but I
honestly don't know how to remedy that.

This commit does TWO things:

1. It introduces a DictSplit transformer, which is the
   map[string]interface{} equivalent to the []interface{} version which
   is Split.
2. It adds MetadataName to both of them, which allows saving the key
   which is split on to metadata. For arrays, this will be the array
   index, starting with 0. For dict, it's the key of the dictionary,
   which is clearly useful.

(Test cases are fixed in second commit)

### DIFF
--- a/transformer/split_test.go
+++ b/transformer/split_test.go
@@ -80,7 +80,8 @@ func TestSplit(t *testing.T) {
 
 	split_path := "data"
 	metadata := transformer.Split{
-		Field: []string{split_path},
+		Field:        []string{split_path},
+		MetadataName: "arrayidx",
 	}
 
 	if err := metadata.Transform(&c); err != nil {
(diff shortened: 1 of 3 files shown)


transformer/split_test.go
```diff
@@ -97,17 +98,100 @@ func TestSplit(t *testing.T) {
...
+	test2 := 1
+	if c.Metrics[0].Data["name"] == "foo" && c.Metrics[1].Data["name"] == "bar" {
+		test1 = 0
+		test2 = 1
+	} else if c.Metrics[1].Data["name"] == "foo" && c.Metrics[0].Data["name"] == "bar" {
+		test1 = 1
+		test2 = 0
+	} else {
+		t.Errorf(`Expected Metrics not present for dict split?`)
+	}
```
sklirg, line 187: I think this warrants a comment about the assignment of test1/test2. 😬

sklirg (approved):
Looks good to me. Would love the little comment I suggested, though.

PR #202, opened 2021-11-22
fix: Allow syslog structured data without hostname

sklirg:
f607809ec0d70780bab8c0612cc2d2059b9228c1 contains a fix and the following commits contain a cleanup/refactor of the whole parser.

The general idea of the refactor was to rework how metrics are parsed, as structured data entries are allowed in multiple forms, while the parser assumed one form and "accepted" the other. This fix, therefore, makes both forms first-class citizens.

The forms are

```
[foo@example key=val]
[bar@example key=val]
```

and 

```
[foo@example key=val][bar@example key=val]
```

and the original parser handled the first one by splitting on newlines but had to "add in" support for multiple entries on the same line. Now, this support is mainline.

### DIFF
--- a/parser/structured_data.go
+++ b/parser/structured_data.go
@@ -165,10 +165,10 @@ func structuredDataParser(bytes []byte, removeEscapedCharsFromResult, stopOnNewM
 	escape := false
 	escapeChars := make([]int, 0)
 	escapeCharsWidth := make([]int, 0)
-	start := 0
-	for width := 0; start < len(bytes); start += width {
+	tokens = 0
+	for width := 0; tokens < len(bytes); tokens += width {
 		var c rune
-		c, width = utf8.DecodeRune(bytes[start:])
+		c, width = utf8.DecodeRune(bytes[tokens:])
 
 		if escape {
 			escape = false
@@ -202,7 +202,7 @@ func structuredDataParser(bytes []byte, removeEscapedCharsFromResult, stopOnNewM
 		if c == '\\' {
 			escape = true
 			if removeEscapedCharsFromResult {
-				escapeChars = append([]int{start}, escapeChars...)
+				escapeChars = append([]int{tokens}, escapeChars...)
 				escapeCharsWidth = append([]int{width}, escapeCharsWidth...)
 			}
 			continue
@@ -217,13 +217,12 @@ func structuredDataParser(bytes []byte, removeEscapedCharsFromResult, stopOnNewM
 		}
 	}
 
-	// Prepare the return values
-	data = bytes[:start]
-	tokens = len(data)
+	// Prepare the return value
+	data = bytes[:tokens]
 
 	for i, escapedChar := range escapeChars {
 		if removeEscapedCharsFromResult {
-			data = []byte(fmt.Sprintf("%s%s", data[0:escapedChar], data[escapedChar+escapeCharsWidth[i]:start]))
+			data = []byte(fmt.Sprintf("%s%s", data[0:escapedChar], data[escapedChar+escapeCharsWidth[i]:tokens]))
 		}
 		tokens += escapeCharsWidth[i]
 	}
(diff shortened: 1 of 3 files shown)


parser/structured_data.go
```diff
@@ -51,53 +51,46 @@ func (sd *StructuredData) Parse(bytes []byte) (*skogul.Container, error) {
 	}, nil
 }
 
-// mnrParseData takes the raw input and parses it
+// parseStructuredData takes the raw input and parses it
 // this takes care of splitting input on newlines etc
 func (sd *StructuredData) parseStructuredData(data []byte) ([]*skogul.Metric, error) {
-	lines := bytes.Split(data, []byte("\n"))
+	lineScanner := bufio.NewScanner(bytes.NewReader(data))
```
KristianLyng, line 70: Ah, but it'ls not technically splitting on lines any more?

parser/structured_data.go
```diff
@@ -51,53 +51,46 @@ func (sd *StructuredData) Parse(bytes []byte) (*skogul.Container, error) {
...
 
-		var metric *skogul.Metric
-		for {
-			canContinue := kvScanner.Scan()
-
+		for kvScanner.Scan() {
 			tag := strings.Trim(kvScanner.Text(), "\u0000")
 			tagValue := strings.SplitN(tag, "=", 2)
 
+			if len(tagValue) == 1 && metric.Metadata["sd-id"] == nil {
```
KristianLyng: wat?

I take it this is some magic structured data detail, but what's sd-id and what's going on here? Some comment, and possibly user-documentation seems reasonable.

parser/structured_data.go
```diff
@@ -191,29 +204,39 @@ func structuredDataParser(data []byte, removeEscapedCharsFromResult bool) (int,
...
+		// in which case we will keep going until
+		// we find a closing bracket.
+		if !stopOnNewMetric && c == ' ' {
 			break
 		}
 	}
 
-	skippedWidth := 0
+	// Prepare the return values
+	data = bytes[:start]
```
KristianLyng: This is a new way of looking at the word "start" :D

KristianLyng (commented):
It looks mostly good to me, mainly language-nitpicks which can be ignored. I would like to see some clarification on sd-id though, either code comment or user documentation or both, because it's somewhat magical. Other than that, I approve.

parser/structured_data.go
sklirg: Well, we kind of are: https://github.com/telenornms/skogul/blob/d311a6bb03335f4704ec0e250ef0bbf7653b79a2/parser/structured_data.go#L174-L178

We just don't split the data on newlines *first*, and *then* start looking for closing `]`'s.

We loop until we find an unescaped `]` or `\n`, it doesn't matter which. And if we get a `]\n` we'll just short-circuit the next iteration and start parsing from the next char if there are more left.

parser/structured_data.go
sklirg: I can refer to the RFC for what the SD-ID is, but this line of code was just moved from here:
https://github.com/telenornms/skogul/commit/35dce7c3ba03e82ad6e4fed0365dd53425d47608#diff-d88400ccea9a24abee535154bbe87f6e872ac0555ca0e765f8315c39c929b1d1L99

parser/structured_data.go
sklirg: Yeah... I started refactoring the names there but I obviously _start_ed in the wrong end... or stopped...

parser/structured_data.go
sklirg: Since parsers don't have configurable fields it's a bit hard to do any more than that... 🤔

KristianLyng: I didn't spot the move, that's on me.

However: Parser _can_ have configurable fields now like all modules, just, this one doesn't. You could, however, mention it in parser/auto.go .  But go ahead and merge if you feel like it, since this isn't a part of the PR.

KristianLyng (approved):

PR #203, opened 2021-11-29
Minor language tweaks in docs and make null-auto

KristianLyng:


sklirg (approved):
LGTM. Just restart the test and hope something doesn't use port 1339.. :)

Issue #159, opened 2020-10-28
Allow ermbedding a stats/count-sender in SenderRef


KristianLyng:
The count-sender does a decent job of providing insight, but if you want to fully utilize it, you need to have a ton of them and manually define all of them.

Instead, I suggest we aim towards implicit injection of it.

When we load config today, we have a SenderRef map. Before we finish config-loading, we can synthesize a count-sender for each SenderRef, set it up with a name that matches the sender, then set up Next to point to the actual sender. This will provide stats for every sender listed, and because we already use SenderRef everywhere, it doesn't require other changes to the code.

In the first iteration, a global config option should be provided that controls the behavior. There are mainly two things to worry about: 1. Enable/disable and 2. What to do with the stats.

In the first version, it would be disabled by default. But the ambition should be to have it enabled by default in future releases, however, that will require a default sender that can expose the stats in an autonomous manner. That's a problem for a later task.

As for what to do with it: In the first version it would require a destination sender and default to "debug". It might seem like a problem that sending stats would use a sender which would then also send stats, but that's surprisingly uproblematic: The count-sender doesn't issue 1 extra send per send, so there is no loop or feedback loop.

The long-term ambition of this feature is to also introduce it for ParserRef and TransformerRef, and learn from it to see how we can do more custom stats efficiently as well. Issue #161 will deal with ParserRef and TransformerRefs.

PR #205, opened 2021-11-30
feat: Stats framework

### LINKED ISSUE #160: Update count-sender to add error-stats
We need more stats from the count-sender, let's start by adding errors.

sklirg:
Initial work at #159, and might supersede and close #160 as well.

Note that this change request only adds the stats *framework*, as well as an "early access" (...) preview for the UDP receiver.

### DIFF
--- a/receiver/auto.go
+++ b/receiver/auto.go
@@ -66,7 +66,7 @@ func init() {
 	Auto.Add(skogul.Module{
 		Name:  "stats",
 		Alloc: func() interface{} { return &Stats{} },
-		Help:  "Periodically ship Skogul stats.",
+		Help:  "Gather internal Skogul metrics and send them on to the specified handler. Metrics gathered depends on modules used, and verbosity and completeness also depends on the modules. Examples of metrics gathered are: parse errors, send errors, number of received messages.",
 	})
 	Auto.Add(skogul.Module{
 		Name:  "stdin",
--- a/receiver/auto.go
+++ b/receiver/auto.go
@@ -64,6 +64,11 @@ func init() {
 		Help:  "Listen for Skogul-formatted JSON on a MQTT endpoint.",
 	})
 	Auto.Add(skogul.Module{
+		Name:  "stats",
+		Alloc: func() interface{} { return &Stats{} },
+		Help:  "Periodically ship Skogul stats.",
+	})
+	Auto.Add(skogul.Module{
 		Name:  "stdin",
 		Alloc: func() interface{} { return &Stdin{} },
 		Help:  "Reads from standard input, one collection per line, allowing you to pipe collections to Skogul on a command line or similar.",
--- a/receiver/stats.go
+++ b/receiver/stats.go
@@ -1,0 +1,112 @@
+/*
+ * skogul, stats receiver
+ *
+ * Copyright (c) 2019-2020 Telenor Norge AS
+ * Author(s):
+ *  - Håkon Solbjørg <hakon.solbjorg@telenor.no>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301  USA
+ */
+
+package receiver
+
+import (
+	"time"
+
+	"github.com/telenornms/skogul"
+)
+
+var statsLog = skogul.Logger("receiver", "stats")
+
+// Stats sends metrics to a HTTP endpoint
+type Stats struct {
+	Handler           *skogul.HandlerRef
+	Interval          skogul.Duration
+	SendEveryInterval bool `doc:"Send stats for every configured interval, even if no new stats are to be reported."` //FIXME: Skogul crashes on sending empty metrics
+	ch                chan *skogul.Metric
+	ticker            *time.Ticker
+	//ch       chan *skogul.Container // FIXME: no *  (?), we should own this // global chan ?
+}
+
+// Starts starts listening for Skogul stats and
+// emits them on the configured interval.
+func (s *Stats) Start() error {
+	if s.Interval.Duration == 0 {
+		statsLog.Debug("Missing interval for stats reporting, defaulting to every 3 seconds")
+		s.Interval.Duration = 3 * time.Second
+	}
+
+	skogul.StatsChan = make(chan *skogul.Metric, 1000)
+
+	s.ch = make(chan *skogul.Metric, 100)
+
+	s.ticker = time.NewTicker(s.Interval.Duration)
+
+	go s.runner()
+
+	for metric := range skogul.StatsChan {
+		s.ch <- metric
+	}
+	return nil
+}
+
+// runner is the function listening for stats and emits
+// them when there is time for it.
+func (s *Stats) runner() {
+	for range s.ticker.C {
+		statsLog.Trace("Time to send skogul stats")
+		// Create a new channel,
+		// consume all on the existing, and then close it
+		ch := s.ch
+		s.ch = make(chan *skogul.Metric, 100)
+		close(ch)
+
+		metrics := make([]*skogul.Metric, 0)
+
+		// Drain the old channel
+		for {
+			statsLog.Trace("Consuming entry from channel..")
+			metric, more := <-ch
+			if !more {
+				break
+			} else if metric == nil {
+				statsLog.Error("Got nil metric on stats channel with more metrics left. Discarding this.")
+				break
+			}
+			metrics = append(metrics, metric)
+		}
+
+		if len(metrics) == 0 && !s.SendEveryInterval {
+			// We have no metrics and we're configured
+			// to *not* ship the metrics if we have none,
+			// so we wait until next tick.
+			statsLog.Trace("Skipping sending metrics since we have none")
+			continue
+		}
+
+		container := skogul.Container{
+			Metrics: metrics,
+		}
+		if err := s.Handler.H.Send(&container); err != nil {
+			statsLog.WithError(err).Error("Failed to send skogul stats")
+		}
+	}
+}
+
+// Verify makes sure all required parameters are set
+func (s *Stats) Verify() error {
+	return nil
+}
--- a/receiver/stats.go
+++ b/receiver/stats.go
@@ -36,7 +36,6 @@ var statsLog = skogul.Logger("receiver", "stats")
 // Stats receives metrics from skogul and forwards it to a handler.
 type Stats struct {
 	Handler  *skogul.HandlerRef
-	Interval skogul.Duration
 	ChanSize uint64
 	ch       chan *skogul.Metric
 	ticker   *time.Ticker
@@ -50,10 +49,6 @@ func (s *Stats) Start() error {
 
 // StartC allows starting Stats with a context.
 func (s *Stats) StartC(ctx context.Context) error {
-	if s.Interval.Duration == 0 {
-		statsLog.Debugf("Missing interval for stats reporting, defaulting to every %d seconds", stats.DefaultInterval)
-		s.Interval.Duration = stats.DefaultInterval
-	}
 	if s.ChanSize == 0 {
 		s.ChanSize = 100
 	}
@@ -63,8 +58,6 @@ func (s *Stats) StartC(ctx context.Context) error {
 	// or we'll have to direct stats to a specific stats instance.
 
 	s.ch = make(chan *skogul.Metric, s.ChanSize)
-
-	s.ticker = time.NewTicker(s.Interval.Duration)
 
 	go s.runner()
 
--- a/receiver/stats.go
+++ b/receiver/stats.go
@@ -1,7 +1,7 @@
 /*
  * skogul, stats receiver
  *
- * Copyright (c) 2019-2020 Telenor Norge AS
+ * Copyright (c) 2021 Telenor Norge AS
  * Author(s):
  *  - Håkon Solbjørg <hakon.solbjorg@telenor.no>
  *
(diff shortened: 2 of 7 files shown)


sklirg:
Starting this as a draft just to get some thoughts and insights.

receiver/stats.go
```diff
@@ -0,0 +1,155 @@
+/*
+ * skogul, stats receiver
+ *
+ * Copyright (c) 2019-2020 Telenor Norge AS
```
KristianLyng: Should say just Copyright (c) 2021

receiver/stats.go
```diff
@@ -0,0 +1,155 @@
...
+var statsDrainCtx, statsDrainCancel = context.WithCancel(context.Background())
+
+// init makes sure that the skogul stats channel exists at all times.
+// Furthermore, it starts a goroutine to empty the channel in the case
+// that the stats receiver is not configured, in which case the chan
+// would end up blocking after it is filled.
+func init() {
+	// Create skogul.StatsChan so we don't have components blocking on it
+	if skogul.StatsChan == nil {
+		skogul.StatsChan = make(chan *skogul.Metric, 100)
```
KristianLyng: Experience suggests that any buffer size like this should be configurable, but since we don't provide global configuration (yet), it's a bit of a moot point, I suppose.

receiver/udp.go
```diff
@@ -95,6 +119,14 @@ func (ud *UDP) Start() error {
 			ud.Threads = 20
 		}
 	}
+	if ud.EmitStats.Duration == 0 {
+		ud.EmitStats.Duration = 10 * time.Second
```
KristianLyng: Hmm, default emit frequency on a module should probably align with default emit frequency on the receiver (e.g. either 3s og 10s, not one of each)?

receiver/udp.go
```diff
@@ -121,3 +153,45 @@ func (ud *UDP) Start() error {
...
+// startStats starts the listening loop for the stats chan
+func (ud *UDP) startStats() {
+	for {
+		switch <-ud.stats.ch {
+		case receive:
+			ud.stats.Generic.Received++
+		case er:
+			ud.stats.Generic.Errors++
+		case sent:
+			ud.stats.Generic.Sent++
```
KristianLyng: I don't really see the benefit of using a channel and a secondary thread here. Wouldn't atomic.AddUint64 for instance do just as well, but with lower overhead? I know channels are fast, but it's difficult to believe they are faster than an atomic +1.

receiver/stats.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	go DrainStats(statsDrainCtx)
+}
+
+// drainStats drains all statistics on the stats channel.
+// If the passed context is cancelled it will stop draining the channel
+// so that a configured stats-receiver can listen on the channel.
+func DrainStats(ctx context.Context) {
```
KristianLyng: I think this should live outside the receiver, along with skogul.StatsChan. As long as skogul.StatsChan exists, it needs to be drained, and that's a fundamental responsibility of the core, and not a module. If I read this right, it shouldn't change any of the principles, just clarify ownership. It might, however, make sense to make a stat/ package (github.com/telenornms/skougl/stat) to contain all of this, since I anticipate a few more helper-functions.

KristianLyng (commented):
The core mechanic of a channel with metrics and draining it feels right to me. It's a generic approach and basically infinitely extensible. There are some implementation nitpicks of course (this is a draft, after all), however the two major thoughts I have are:

1.I just don't see the value of skogul.Stats. It might be OK if modules are expected to _only_ or at least mainly provide a similar set of stats, but I don't see that as even remotely realistic. For the HTTP receiver we want to count individual error codes, for example. It also makes it less intuitive in the individual modules what's going on. And at the end of the day, I just feel like it's unnecessary. Just setting up a Metric in each module shouldn't be a big deal, should it? Counter arguments are welcome.

2. I feel fairly strongly that Skogul should work if we just delete any random module from the source tree - in theory. Sure, you won't get the benefit of that module and there are some reasonable exceptions, but there should be no core functionality "hidden away" in modules. As such, the part of the receiver code that takes care of draining the channel should, imo, live outside the receiver itself, and the receiver should instead hook into that. That way, we could re-implement the receiver and have two different variations of it living in parallel in the code tree. From what I can see, this shouldn't represent a major obstacle or big change - mainly just copy out the init-related stuff?

Other than that, I think we need to get the reverse map lookup I've mentioned before in place before we merge this, since there's no way to distinguish different modules now, we can consider that a different project/feature/branch though.

receiver/stats.go
sklirg: Damnit. I remember changing this more than once during my copy/paste endeavours. I'll fix.

receiver/udp.go
sklirg: I haven't done any benchmarking, but it sounds possible that it has a lower overhead. I'm not heavily invested in either approach, so I'll swap it. Two of my "pros" from the channel approach was that it is possible to create an "enum" type which makes it very easy to read how stats are generated/incremented, compared to the more verbose atomic.AddUint64() (`ud.stats.ch <- received` vs. `atomic.AddUint64(ud.stats.Received, 1)`). The other is that there is a (symbolic?) separation of concerns between who counts the stats and who generates them, in different goroutines. But that might add more overhead. 🤷‍♂️

sklirg:
> 1. I just don't see the value of skogul.Stats. It might be OK if modules are expected to _only_ or at least mainly provide a similar set of stats, but I don't see that as even remotely realistic. For the HTTP receiver we want to count individual error codes, for example. It also makes it less intuitive in the individual modules what's going on. And at the end of the day, I just feel like it's unnecessary. Just setting up a Metric in each module shouldn't be a big deal, should it? Counter arguments are welcome.

I was thinking of the Stats struct for mostly skogul-only internal stats (this is very meta), i.e. metrics for receiver->parser, parser->sender, sender->. The receiver can fail to receive data (TCP connection closed in the middle of a read), the parser can be unable to parse the data, and the sender can be unable to send the metrics. This would _maybe_ reduce some boilerplate, or at the very least define a struct of stats that should be implemented in every module.

We can start without this, but if we feel the need for it or that it would actually help in the future, we can of course add it back.

> 2. I feel fairly strongly that Skogul should work if we just delete any random module from the source tree - in theory. Sure, you won't get the benefit of that module and there are some reasonable exceptions, but there should be no core functionality "hidden away" in modules. As such, the part of the receiver code that takes care of draining the channel should, imo, live outside the receiver itself, and the receiver should instead hook into that. That way, we could re-implement the receiver and have two different variations of it living in parallel in the code tree. From what I can see, this shouldn't represent a major obstacle or big change - mainly just copy out the init-related stuff?

Agreed. I've moved it to a separate stats package. This makes receiver/stats.go depend on the stats package, but since that is more of a "core module" than a receiver/sender module, I feel this is acceptable.

stats/main.go
```diff
@@ -0,0 +1,75 @@
...
+		case <-StatsChan:
+			continue
+		case <-ctx.Done():
+			statsLog.Debug("Stopping stats drain. Stats are being consumed.")
+			return
+		}
+	}
+}
+
+// DefaultInterval is the default interval used for sending stats.
```
KristianLyng: Can we move this to the top of the file?

stats/main.go
```diff
@@ -0,0 +1,75 @@
+/*
```
KristianLyng: Eh, can we name the file anything except main? In my head, main.go suggest a main-method (e.g.: cmd/skogul/main.go). Might just be me, but humor me? :D

stats/main.go
```diff
@@ -0,0 +1,75 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301  USA
+ */
+
+package stats
```
KristianLyng: Obviously reasonable that it's omitted in a draft, but the added bonus of having this a separate package is that we now have a natural place to document how the stats subsystem works.

KristianLyng (commented):
Ok, I really like this now, it's just so simple, yet powerful.

I figure we'll expand the stats package with any relevant helper-functions if we realize there's a need (similar to what you suggest).

I basically only have nitpicks here. I didn't re-review the details of the receiver since I assume it isn't considerably changed. And I only skimmed the UDP sender, and it looked fine.

One such nitpick: Now that stats is a package, prefixing exported objects with Stats is redundant, e.g: stats.StatsChan should be stats.Chan or similar. 

Other than that, some very rudamentary code-documentation is really the only thing I'm missing, and then I'd be ready to start using this "everywhere" more or less and see how it pans out.

receiver/auto.go
```diff
@@ -63,6 +63,11 @@ func init() {
 		Alloc: func() interface{} { return &MQTT{} },
 		Help:  "Listen for Skogul-formatted JSON on a MQTT endpoint.",
 	})
+	Auto.Add(skogul.Module{
+		Name:  "stats",
+		Alloc: func() interface{} { return &Stats{} },
+		Help:  "Periodically ship Skogul stats.",
```
KristianLyng: This definitely needs to be a bit more verbose. 

Suggestion: "Gather internal Skogul telemetry and ship it. What telemetry is gathered depends on the modules used. Some modules send no stats at all, while others might log things such as number of parse errors or number of sent messages."

receiver/stats.go
```diff
@@ -0,0 +1,126 @@
...
+		statsLog.Debugf("Missing interval for stats reporting, defaulting to every %d seconds", stats.DefaultInterval)
+		s.Interval.Duration = stats.DefaultInterval
+	}
+	if s.ChanSize == 0 {
+		s.ChanSize = 100
+	}
+
+	// XXX: we shouldn't allow multiple stats receivers instantiated probably
+	// because they'll just steal the stats from each other.
+	// or we'll have to direct stats to a specific stats instance.
```
KristianLyng, line 58: I see several ways of detecting this (e.g.: global module-specific variable). For now, I suggest just making an issue for it and we can get back to it so we can get this merged.

receiver/stats.go
```diff
@@ -0,0 +1,126 @@
...
+
+	s.ticker = time.NewTicker(s.Interval.Duration)
+
+	go s.runner()
+
+	stats.StatsDrainCancel()
+
+	for metric := range stats.Chan {
+		if len(s.ch) >= cap(s.ch) {
+			statsLog.Debug("Dropping stats because the channel is full")
```
KristianLyng, line 68: Ah, you do realize the irony of this not being logged as stats, right? :D But that can be a different PR/Issue I suppose.

receiver/stats.go
```diff
@@ -0,0 +1,126 @@
...
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
+	return nil
+}
+
+// runner is the function listening for stats and emits
+// them when there is time for it.
+func (s *Stats) runner() {
```
KristianLyng, line 83: Ok, I know I've approved this earlier and I think it's fine the way it is, but I think what we want is a different logic.

Right now, metrics are sent by a module, then artificially delayed for no apparent benefit. If the stats were actually modified up until this interval, it would make sense, but right now, I just don't know what the benefit of this algorithm is as opposed to just sending everything as it is received? Sure, it ends up batching more metrics together, which is more performant, but we have a batch sender that'll do that even better.

I feel like we had some idea here for this logic that slipped our mind someway along the road to implementation.

receiver/udp.go
```diff
@@ -49,9 +52,19 @@ type UDP struct {
 	PacketSize   int               `doc:"UDP Packet size note: max. UDP read size is 65535"`
 	FailureLevel string            `doc:"Level to log receiver failures as. Error, Warning, Info, Debug, or Trace. (default: Error)"`
 	Buffer       int               `doc:"Set kernel read buffer. Default is kernel-specific. Bumping this will make it easier to handler bursty UDP traffic."`
+	EmitStats    skogul.Duration   `doc:"How often to emit internal skogul stats for this receiver"`
```
KristianLyng: What's the default? I also think we want a chapter in the hard-coded part of the manual on internal stats, but that can be a separate PR/Issue.

stats/chan.go
```diff
@@ -0,0 +1,81 @@
...
+// StatsChan is a channel which accepts skogul statistic as a skogul.Metric
+// By configuring the stats receiver, this channel is drained and sent on to
+// the specified handler.
+var Chan chan *skogul.Metric
+
+// StatsDrainCtx and StatsDrainCancel are the context and cancel functions
+// for the automatically created stats.StatsChan.
+// If a skogul stats receiver is configured, StatsDrainCancel MUST be called
+// so that statistics are not discarded.
+var StatsDrainCtx, StatsDrainCancel = context.WithCancel(context.Background())
```
KristianLyng: Same comment with name as I had on the channel earlier. Let's not talk about stats.StatsDrainCtx when we could be talking about stats.DrainCtx :D

KristianLyng (changes requested):
So it looks ok.

I made a lengthy comment I think needs to be addressed on the core mechanic of the receiver though. I missed it during previous reviews, and since it's a receiver and a module, it means we CAN fix it later, but I just don't see what the value of the current round-about way of adding an interval is when we can just ... not have an interval and read the metrics as they are received. 

I'm jolting this as a "Request changes", mainly because the discussion should be resolved one way or an other, and a few of the other minor comments too should be addressed.

receiver/stats.go
sklirg: My main reasoning for the interval is to keep sending stats even if no new stats have occurred.

Another option is to send stats as soon as they happen *and* periodically. What do you think?

sklirg: Slightly never mind my previous comment. We have two intervals configured: one in the module sending stats and one in the stats-receiver. I agree that we only need one of these. I think it makes the most sense to remove the one here, in "runner", and allow modules to ship stats on their own interval - and skogul will ship them on immediately. If wanted, batchers or other senders can be used to achieve the same benefit.

receiver/stats.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/telenornms/skogul"
+	"github.com/telenornms/skogul/stats"
+)
+
+var statsLog = skogul.Logger("receiver", "stats")
+
+// Stats receives metrics from skogul and forwards it to a handler.
+type Stats struct {
+	Handler  *skogul.HandlerRef
+	Interval skogul.Duration
```
KristianLyng: Interval is no obsolete.

KristianLyng (commented):
Ok, with the exception that Interval is still left even though it doesn't do anything this looks good to me :D

KristianLyng (approved):

PR #206, opened 2021-12-01
feat(testing): Try to find a new port if the specified one is in use

sklirg:
This should fix some of the tests we currently have where we try to listen on a port that might be unavailable in a CI environment.


KristianLyng (approved):
Nifty :D

PR #207, opened 2021-12-02
Add optional burner to the batcher

KristianLyng:
The burner sender can be used to "burn" metrics if the batcher would
otherwise end up blocking. The idea is to use this in scenarios where
you have multiple senders, where some is more important than others, or
you have alternate means of handling bursts of data (e.g.: queue it?).

I think the core mechanic is pretty elegant if I may say so myself, and
it's almost entirely unobtrusive. One could argue that the use of a
pointer to a channel is a bit intricate, I suppose.

I've tested various minor variations of this and seen no significant
improvement on the original. I'm also adding example configuration that
demonstrates the use. Run two skogul instances, one on each config in
docs/examples/burn-test. Try enabling/disabling the burner to see
significant performance differences.

### DIFF
--- a/docs/examples/burn-test/tester_to_udp.json
+++ b/docs/examples/burn-test/tester_to_udp.json
@@ -1,0 +1,34 @@
+{
+  "receivers": {
+    "test": {
+      "type": "test",
+      "handler": "json",
+      "metrics": 10,
+      "values": 10,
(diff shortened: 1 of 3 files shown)


sklirg (approved):
💯 branch name

PR #215, opened 2021-12-07
Add CopyFromData to metadata transformer

### LINKED ISSUE #201: Metadatatransformer: Option to _not_ delete original with ExtractFromData
Basically CopyFromData in addition to ExtractFromData is needed when a value is wanted as both data and metadata. (e.g.: Both filtering and graphing)

KristianLyng:
So this basically duplicates what ExtractFromData does. Because of that,
it also "marks" ExtractFromData as obsolete, but only in documentation.
This can probably be discussed.

It's simple enough though: Copy fields from data to metadata, optionally
rename them and optionally delete the original.

Fixes #201

### DIFF
--- a/transformer/metadata.go
+++ b/transformer/metadata.go
@@ -30,12 +30,22 @@ import (
 	"github.com/telenornms/skogul"
 )
 
+// SourceDestination provides a source and destination key, and the option
+// to delete the source. At this writing, it is only used to copy from the
+// data-section to metadata, but it's left intentionally generic.
+type SourceDestination struct {
+	Source      string `doc:"Name of the source field"`
+	Destination string `doc:"The destination name/field. If left blank/undefined, the source name will be used as a destination name."`
+	Delete      bool   `doc:"Set to true to delete the original. Default is to leave the original."`
+}
+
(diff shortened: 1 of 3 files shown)


KristianLyng:
Btw...

Fixes #201

transformer/metadata.go
```diff
@@ -30,12 +30,22 @@ import (
...
+	Delete      bool   `doc:"Set to true to delete the original. Default is to leave the original."`
+}
+
 // Metadata enforces a set of rules on metadata in all metrics, potentially
 // changing the metric metadata.
 type Metadata struct {
 	Set             map[string]interface{} `doc:"Set metadata fields to specific values."`
 	Require         []string               `doc:"Require the pressence of these fields."`
-	ExtractFromData []string               `doc:"Extract a set of fields from Data and add it to Metadata. Removes the original."`
+	ExtractFromData []string               `doc:"Extract a set of fields from Data and add it to Metadata. Removes the original. Obsolete, will be removed. Use CopyFromData instead."`
```
sklirg, line 47: Should we have a warning for this deprecation in the Verify() function?

sklirg (approved):
LGTM.

PR #217, opened 2021-12-09
Stats for the HTTP receiver

Eiriktaa:
@sklirg Introduced me to the codebase and helped me getting started adding some stats to the HTTP receiver.

This is branched from #213


sklirg (approved):
Assuming #213 is merged, I think this looks good, but I'd like input from @KristianLyng on it as well.

KristianLyng:
I think this looks good too, thanks a bunch for the patches (and sorry for the slow response :D)

Issue #228, opened 2022-04-29
Adding "encoder" modules


KristianLyng:
Today, receivers are agnostic to the data encoding format. E.g.: The HTTP receiver can decode json, influx, protocol buffers, m&r etc, and all the other receivers can do that as well, since it's coupled with the handler.

However, the same can not be said for senders. Most of the time this isn't a big deal, since specific senders typically require small tweaks - the influx sender might require specific HTTP headers for sending a token to authenticate for example.

However, it's becoming increasingly obvious that being able to separate encoding from transport would be beneficial. So as such, a new module type is required, the inverse of parsers (coincidentally, maybe we should rename parsers to "decoders"?). 

I don't think every sender should be required to use an encoder - it makes no sense at all to implement an SQL sender using a generic encoder for example, and by and large, sender modules should do what you expect them to do by default, with no encoder set. But this change should at least reduce code re-use in the HTTP senders(plural - http, influx and hec all use similar code).

This also ties in nicely with #227 - being able to send GOB data over HTTPS, but also UDP could make sense - use UDP to duplicate data to test instances, which would isolate the sending skogul from issues with the receiving skogul.

Great care needs to be taken to avoid further overcomplicating the configuration though.

KristianLyng:
I've added this support, but so far there's only one encoder which makes this completely worthless.

This issue should track adding of the following encoders, which will likely require unifying some logic in the HTTP sender:

- [ ] Influx line protocol encoder
- [ ] HEC encoder

KristianLyng:
Ok, I ran into a conundrum today. The Kafka sender has implicit support for batching, that means the metrics[] logic is superfluous. I ended up adding a EncodeMetric method to the Encoder interface, but I'm still not convinced this is a universal thing, and it means there's a miss-match between encoder and parser.

 The approach I used will "flatten" the container and transmit each metric as independent messages, which presumably makes things easier for other consumers, but it also mandated that I made a "JSONMetric"/"SkogulMetric" parser to match it, and now we have a situation where the JSON encoder can be used to send data with the HTTP sender, and then parsed by the JSON parser through a HTTP receiver, but you CAN'T use the JSON parser to parse messages sent by over Kafka, using the same encoder, because it's up to the sender which method to use.

I don't think there's an obvious solution here, and currently I'm leaning towards just seeing how this plays out as we extend the encoder-concept.

PR #240, opened 2022-06-28
Implement GOB (and avro) encoder and parser

### LINKED ISSUE #227: GOB parser and encoder
While json-encoded HTTPS is versatile, it is not particularly efficient. Protocol buffers are not self describing, so they are not well suited for inter-skogul communication either. 

However, gobs ( https://pkg.go.dev/encoding/gob - https://go.dev/blog/gob ) should be the right fit for skogul-to-skogul traffic. It can be transported with any of the existing senders.

Gobs are only selfdescribing within the same stream, so a new stream will have to be created for each container, this reduces the effectiveness somewhat, but is required to avoid two-way communication between encoder and decoder to determine when to re-send the headers/"reboot" the encoder after a decoder restart/loss of state. 

In short: Whenever the encoder is called, a new bytes.Buffer and gob.NewEncoder() needs to be set up. We should probably add benchmarks to both this and the json encoder to compare. Regardless of performance, I believe it's worthwhile because with JSON we have information loss due to lack of typing: We can't determine if '0' is an integer or a floating point. 

EncodeMetric() and Encode() can be identical.

RoshiniNara:
fixes #227 and #228


encoder/auto.go
```diff
@@ -1,9 +1,9 @@
 /*
  * skogul, Encoder automation
  *
- * Copyright (c) 2019-2020 Telenor Norge AS
+ * Copyright (c) 2022 Telenor Norge AS
  * Author(s):
- *  - Kristian Lyngstøl <kly@kly.no>
```
KristianLyng, line 6: Please don't remove me :D (Adding yourself is fine)

encoder/auto.go
```diff
@@ -54,4 +54,12 @@ func init() {
 		Help:     "Encodes the standard Skogul JSON format with indentation.",
 		AutoMake: true,
 	})
+	Auto.Add(skogul.Module{
+		Name:     "GOBskogul",
```
KristianLyng: I think just "gob" makes more sense than "gobskogul", in which case the entire line that reads Aliases: can be removed so it just says Name: "gob"

encoder/gob_test.go
```diff
@@ -0,0 +1,128 @@
...
+	old1 = "\x1d"
+	old2 = "\x0e"
+	old3 = "main"
+	n = 1
+	new1 = "\x1f"
+	new2 = "\x10"
+	new3 = "skogul"
+	result1 = bytes.Replace([]byte(origstring1), []byte(old1), []byte(new1), n)
+	result2 = bytes.Replace([]byte(result1), []byte(old2), []byte(new2), n)
+	result3 = bytes.Replace([]byte(result2), []byte(old3), []byte(new3), n)
```
KristianLyng, line 82: If I read this correctly, you're replacing three things in the parsed gob data before comparing with the encoded data. I don't understand why this is necessary, shouldn't they be the exact same? Or is this because the testdata was generated with a different tool? Either way, at least a comment is needed explaining it.

encoder/gob_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	sorig_trim := string(result3)
+	snew := string(b)
+
+	//sorig_trim := strings.TrimSpace(string(result3))
+	//snew_trim := strings.TrimSpace(snew)
```
KristianLyng: Probably want to remove these if they aren't needed.

parser/auto.go
```diff
@@ -81,4 +81,18 @@ func init() {
...
+	Auto.Add(skogul.Module{
+		Name:     "GOB",
+		Aliases:  []string{"gob"},
+		Alloc:    func() interface{} { return &GOB{} },
+		Help:     "Parse GOB",
+		AutoMake: true,
+	})
+	Auto.Add(skogul.Module{
+		Name:     "GOBmetric",
+		Aliases:  []string{"gobmetric"},
```
KristianLyng: So far every module has a lower case Name, if there aren't any relevant aliases it's better to just drop the entire Aliases-line and just have a lower-case name. If we want upper/lower case we should just make it case insensitive in the configuration parser instead.

parser/auto.go
```diff
@@ -81,4 +81,18 @@ func init() {
...
+		Alloc:    func() interface{} { return &GOBMetric{} },
+		Help:     "Parses GOBMetric.",
```
KristianLyng: The Help strings probably needs to be a bit more extensive since this is the only user-facing documentation. Not saying all the other Help-texts are perfect either though... I suggest something along the lines of "Parse a GOB-encoded Skogul Container. GOB is a go-specific encoding, useful for inter-Skogul communication." for the regular gob parser, and similar, but "Parse a single GOB-encoded Skogul Metric." for the metric version. Something similar for the encoder, so the user has at least some idea what this does without looking up what GOB is... (Texts are just suggestions).

parser/gob.go
```diff
@@ -0,0 +1,33 @@
+package parser
```
KristianLyng, line 23: Please add the copyright header at the top of the file.

KristianLyng (changes requested):
Looks mostly good, just some minor issues I've commented on. The biggest - beyond the test failures - is that config/https_gob_to_stdout.json and config/tester_to_https_gob.json needs to be moved. Configuration examples live in docs/examples and since these two are related, I suggest adding a new directory called docs/examples/avro/ and putting both in there instead, ideally a short README.rst explaining what they do - something short is fine, see for example docs/examples/kafka/README.rst for a short example.

Some background: Everything in config/ is the actual code responsbile for parsing and applying configuration files, not meant for *storing* actual configuration. Conversely, everything in docs/ is distributed when you install Skogul on a machine, and every configuration file in docs/examples/ will be automatically tested for validity (i.e.: testet if it parses, not actually run) on "make check".

parser/schema/avro_schema
```diff
@@ -0,0 +1,39 @@
...
+								"values": ["string", "int", "long", "float", "double" ]
+							}
+						}
+						
+					]
+				}
+			}
+		}
+	]
+}
```
sklirg: This is the same file as `encoder/schema/avro_schema` as far as I can tell, is there a reason for it being duplicated, or could we have only one file, but referenced for purposes from both encoder and parser? That way we don't have any "Oh I forgot to update the other schema file"-mistakes :)

encoder/auto.go
RoshiniNara: Fixed

encoder/gob_test.go
RoshiniNara: The test gob data is generated using a separate program and there are differences in 3 locations compared to the encoded data.
These differences are fixed by replacing the differences with exact matches before the comparison. I added the comments in the code.

encoder/gob_test.go
RoshiniNara: I removed the line.

parser/auto.go
RoshiniNara: Fixed.

parser/auto.go
RoshiniNara: Added the helper strings as per your suggestions.

parser/schema/avro_schema
RoshiniNara: I  will follow the recommendation @KristianLyng  here.

encoder/avro.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/telenornms/skogul"
+)
+
+type AVRO struct {
+	Schema string
+}
+
+func (x AVRO) Encode(c *skogul.Container) ([]byte, error) {
+	b, _ := os.ReadFile(x.Schema)
+	schema := avro.MustParse(string(b))
```
KristianLyng: This part needs to be inside sync.Once, similar to how it's done in transformer/timestamp.go for instance. Otherwise Skogul will open the schema on disk for every single AVRO packet, which could be several thousand per second.

It ALSO needs to check the return code of os.ReadFile(), so change `b, _ :=` to `b, err := ` and check if err is nil or not.

KristianLyng (changes requested):
Moving config examples

config/https_gob_to_stdout.json and config/tester_to_https_gob.json needs to be moved. To do this, you need to tell git about it. 

1. Make a directory, docs/examples/gob
2. Run "git mv config/https_gob_to_stdout.json docs/examples/gob" and "git mv config/tester_to_https_gob docs/examples/gob"
3. git commit and push

Schema

I think it makes sense for the AVRO-schema you provide, which is technically meant as an example, to live in docs/examples/avro. 

So I suggest you move it in there, which will also take care of Håkon's input, and then update the references in the test cases.

Failing tests

The tests fail right now mainly because of #245 which is unrelated to this pull request. As I mention in that issue, I think it makes sense if you upgrade to go 1.19, run make fmtfix, and attach it to this PR so we can see  tests working.

Final thoughts

Things look good in general. Most of the feedback I have are nitpicks about style and general code organization, not about the implementation.

What I suggest are:

1. Move the files as mentioned above
2. Run "make fmtfix" for go 1.19 to make the tests pass
3. Create two new issues here on github: one for timestamps not working in avro, and one for introducing sync.Once to the avro encoder/decoder. 

This will let you finish this pull request, then we can work on the other issues separately as smaller problems.

PR #244, opened 2022-08-02
chore(transformer/timestamp): Don't try to re-parse the timestamp every time it's used

sklirg:
Also reduce log-level to debug for case-match of timestamp strings.

This printed a log line in `-loglevel info` every time a timestamp was parsed with a custom format. It might be useful to see the message once, but not for every transform. I made it so the timestamp format is set once during init, not for every transform, and reduced the log-level to not post it in info-logs.

### DIFF
--- a/transformer/timestamp.go
+++ b/transformer/timestamp.go
@@ -61,6 +61,7 @@ func (config *Timestamp) Transform(c *skogul.Container) error {
 		if config.Format == "" {
 			config.Format = time.RFC3339
 		}
+		config.Format = parseTimestamp(config.Format)
 	})
 
 	for i, metric := range c.Metrics {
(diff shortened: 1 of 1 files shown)


transformer/timestamp.go
```diff
@@ -61,6 +61,7 @@ func (config *Timestamp) Transform(c *skogul.Container) error {
 		if config.Format == "" {
 			config.Format = time.RFC3339
 		}
+		config.Format = parseTimestamp(config.Format)
```
KristianLyng: So this overwrites the original config.Format which is user-provided. Should this be an unexported one instead? (ps: the rest looks good though)

KristianLyng (commented):
Mostly looks good to me, just curious about the Format overwrite.

transformer/timestamp.go
sklirg: It's because we provide a convenience function to support short formats for standards (iso8601/RFC3339), but we need to set the actual format. It's a string constant from the time package, so we just set that. So it set the appropriate format if someone configured it by setting the format to "rfc3339" or "iso8601" instead of the full format string, as those are not valid formats.

sklirg: I can make it an un-exported field on the struct instead, yes. 😊

KristianLyng: I think that's marginally cleaner, since otherwise if we want to re-export the config, it will have changed. It's not like it wont change anyway, so this is probably a bit of a weird nitpick...

sklirg: Done 👍

KristianLyng:
You just hit #245 as well... It's just Go 1.19 being mean, since we're working on it, please just leave the 1.19-related formating issues out of this PR to avoid duplication :D

sklirg:
Rebased on primary to get the formatting fixes 👍

PR #246, opened 2022-08-10
chore(fmt): `make fmtfix`

### LINKED ISSUE #245: Update formating for go 1.19 rules
As revealed in test-cases in #240 and as seen in https://go.dev/doc/go1.19 , gofmt has changed. This is a trivial issue, but we need to run gofmt / make fmtfix with Go 1.19.

See https://cloud.drone.io/telenornms/skogul/819/1/3 - this test failure is unrelated to the actual commit being tested, because we just use a genering golang docker-image, which has reasonably been bumped to go 1.19.

I suggest that #240 just incorporates the fix since it's needed to make the checks pass anyway. 

In other words: upgrade to go 1.19, run "make fmtfix", then commit and update the pull request.

sklirg:
Fixes #245

### DIFF
--- a/common.go
+++ b/common.go
@@ -40,7 +40,6 @@ This is needed because the math package doesn't handle 32-bit variants.
 
 The pattern is:
 
-
 seeeeeee e0000000 00000000 00000000
 01234567 01234567 01234567 01234567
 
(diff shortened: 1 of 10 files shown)


KristianLyng:
We ended up fixing this accidentally in #240 so I'm closing this one.

PR #248, opened 2022-08-25
feat(splunk): Send metadata as fields to Splunk HEC

### LINKED ISSUE #242: Metadata configured in Splunk sender is not sent to Splunk
Sending events to Splunk using the Splunk sender omits all metadata-fields, except specifically configured fields like "index" or "host" which can only be configured directly in the Splunk sender.

Source: 
* https://github.com/telenornms/skogul/blob/v0.16.0/sender/splunk.go#L62-L93
* specifically https://github.com/telenornms/skogul/blob/v0.16.0/sender/splunk.go#L86

Where the last code-source explicitly sets the event data to be _only_ the `.data` field from the metric.

---

Potential solution(s):

Splunk doesn't really seem to care about separating data/metadata, so I think the easiest approach would be to merge the two and send all values in a single "event".

Another option is to send the metadata struct as a separate field in the event, as Splunk supports nested objects (see the "kubernetes" object on line 2 of the event):
image

sklirg:
Following the documentation from Splunk HEC about field extractions
during indexing: https://docs.splunk.com/Documentation/Splunk/9.0.0/Data/IFXandHEC#Add_a_fields_property_at_the_top_JSON_level

Before this change, fields in Metadata are silently ignored, and only fields in Data will be sent to splunk.

Fixes #242.

### DIFF
--- a/sender/splunk.go
+++ b/sender/splunk.go
@@ -57,6 +57,7 @@ type splunkEvent struct {
 	Index  string                 `json:"index,omitempty"`
 	Source string                 `json:"source,omitempty"`
 	Event  map[string]interface{} `json:"event"`
+	Fields map[string]interface{} `json:"fields,omitempty"`
 }
 
 // prepare converts a skogul container into the appropriate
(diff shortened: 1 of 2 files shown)


KristianLyng:
Looks good to me, though I have no test setup. I assume this is tested against splunk?

sklirg:
Yes, I've tested using this config (on a local build). Furthermore, "worst case", all that happens is that more data is sent to Splunk - today, all data in "metadata" is silently discarded.. :)

KristianLyng:
Sounds good, and we're all about "more data to splunk" here anyway.

PR #249, opened 2022-08-25
feat(logs): Support configuring skogul to log as JSON

sklirg:
I only added support for JSON, and it will default to "auto" (... no matter what you type other than 'json').
I haven't heard of any request for forcing text mode or color mode, and assume that it works as people expect.
It should be pretty simple to add support for other formats as well if needed.


KristianLyng:
merci :D

Issue #252, opened 2022-09-01
USP protobuf support


KristianLyng:
We need to support USP, specifically:

https://github.com/BroadbandForum/usp/blob/master/specification/usp-record-1-1.proto

and

https://github.com/BroadbandForum/usp/blob/master/specification/usp-msg-1-1.proto

This will be similar to Juniper telemetry, but following the "1-2-many" rule of scaling, I figure we can do one more one-off for this protobuf support. I don't expect frequent updates here, and hopefully this is cleaner than juniper.

On a side note, we should rename the Juniper support from the missleading "protobuf", deprecate the "protobuf" and "telemetry" name in favor of the more accurate vendor-specific name(s), I suggest: protobuf-usp for this.

We have to support this as a parser, encoder is very much optional.

KristianLyng:
Fixed in #254

PR #254, opened 2022-09-13
Add protobuf-usp parser

kamilernerd:
https://github.com/telenornms/skogul/issues/252

As mentioned in the issue. The previous protobuf files have their names changed to a more vendor-specific name. Also usp parser has been added to parser/auto.go.


parser/protobuf_usp.go
```diff
@@ -0,0 +1,152 @@
...
+	"fmt"
+	"sync"
+	"sync/atomic"
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/telenornms/skogul"
+	"github.com/telenornms/skogul/gen/usp"
+)
+
+type P struct {
```
KristianLyng, line 15: Considering this is global to all parsers, I'd prefer a slightly more descriptive name, e.g.: USP or something like that.

parser/protobuf_usp.go
```diff
@@ -0,0 +1,152 @@
...
+	container.Metrics[0] = &recordMetric
+	return &container, err
+}
+
+// Unmarshals []byte into a protoc generated struct and returns it
+func (p *P) getUspRecord(d []byte) (*usp.Record, error) {
+	unmarshaledMessage := &usp.Record{}
+	if err := proto.Unmarshal(d, unmarshaledMessage); err != nil {
+		atomic.AddUint64(&p.stats.ParseErrors, 1)
+		return nil, fmt.Errorf("Failed to unmarshal protocol buffer. Error: %w", err)
```
KristianLyng: Minor nitpick (which we haven't always done right): start error messages with lower case, skip "Error: " and just say "failed to unmarshal protocol buffer: %w" - it avoids having nested error messages with seven instances of the word Error.

parser/protobuf_usp.go
KristianLyng: Given that this is global to all parsers, I'd prefer a slightly more descriptive name, e.g.; USP or something like that.

parser/auto.go
```diff
@@ -67,6 +67,13 @@ func init() {
 		Help:     "Parse Juniper telemetry in the form of protocol buffers. Typicially combined with the UDP receiver.",
 		AutoMake: true,
 	})
+	Auto.Add(skogul.Module{
+		Name:     "protobuf_usp",
+		Aliases:  []string{"usp"},
+		Alloc:    func() interface{} { return &P{} },
+		Help:     "Parse Usp telemetry in the form of protocol buffers. Typicially combined with the UDP receiver.",
```
KristianLyng: So far we haven't got a single use case where it is combined with UDP though. Also: It should probably mention how it's an USP message contained by an USP record, since I for one have no idea whether this is the norm, or just what our hardware happen to do.

parser/protobuf_usp.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// Unmarshals []byte consisting of the record payload into
+// a protoc generated struct and returns it
+func (p *P) getRecordMsgPayload(payload []byte) (*usp.Msg, error) {
+	msgPayload := &usp.Msg{}
+
+	if err := proto.Unmarshal(payload, msgPayload); err != nil {
+		atomic.AddUint64(&p.stats.ParseErrors, 1)
+		return nil, fmt.Errorf("Failed to unmarshal payload. Error: %w", err)
```
KristianLyng: Same nitpick as above re: capital letters and "Error".

parser/protobuf_usp.go
```diff
@@ -0,0 +1,152 @@
...
+	jsonData, err := p.extractJSON(payload.GetBody().GetRequest().GetNotify().GetEvent().GetParams()["Data"])
+
+	if err != nil {
+		return nil, err
+	}
+
+	jsonMap["event"] = payload.GetBody().GetRequest().GetNotify().GetEvent().GetObjPath()
+	jsonMap["event_type"] = payload.GetBody().GetRequest().GetNotify().GetEvent().GetEventName()
+	jsonMap["subscription_id"] = payload.GetBody().GetRequest().GetNotify().GetSubscriptionId()
+	jsonMap["event_parameters"] = jsonData["Report"]
```
KristianLyng: I think some of this might be considered metadata? Except jsonData["Report"]? Let's take a look at some real-life data to evaluate before deciding.

parser/protobuf_usp_test.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		t.Fatalf("unable to read protobuf packet file: %v", err)
+	}
+	if n == 0 {
+		t.Fatalf("read 0 bytes from protobuf packet file....")
+	}
+	return b[0:n]
+}
+
+func TestParseFile(t *testing.T) {
```
KristianLyng: Keep in mind that the scope is for all parser tests, so most exported names should probably have USP in them to avoid future conflicts. (Not that we have been all too strict about this thus far). E.g.: TestUSPParseFile. It also makes it easier when running tests : go test -v -run .*USP.*

KristianLyng (changes requested):
Looks pretty good to me, some minor nitpickks mainly - I am aware that some of the nitpicks I mention here are also present in existing code, as a result of us having learned (specially when it comes to log messages).

Before I accept it, though, I'd like to see some real world usage (which we're planning to do today anyway) to see if we've hit the right metadata/data separation and how the data actually looks.

I also think an example configuration would make sense, but let's hold off on that until we know what the telco standard for transmission of these messages are. We could easily do that as a separate issue anyway.

KristianLyng:
Some more comments:

1. We can leave handling of time out of the issue. USP doesn't specify time at all, so it will be up to the user to either use the now-transformer to set current time to timestamp, or an other timestamp extraction transformer.
2. Comments should follow https://go.dev/doc/comment - particularly the part under "Funcs".

PR #261, opened 2022-09-22
Reduce log noise and remove skogul.Errors

KristianLyng:
I've removed skogul.Error and reduced some log noise. I'd like a review because:

1. In case I made some trivial mistakes - features should be mostly the same
2. Is this generally sensible, it does reduce the logrus fields somewhat, but not dramatically, and general readability is better.

### DIFF
--- a/parser/mnr.go
+++ b/parser/mnr.go
@@ -26,6 +26,7 @@ package parser
 import (
 	"bufio"
 	"bytes"
+	"fmt"
 	"strconv"
 	"strings"
 	"sync"
@@ -99,8 +100,7 @@ func (mnr *MNR) mnrParseData(data []byte) ([]*skogul.Metric, error) {
 		metrics = append(metrics, metric)
 	}
 	if len(metrics) == 0 {
-		mnrLog.WithField("lines", len(lines)).Warnf("MNR parser failed to parse any of the %d lines", len(lines))
-		return nil, skogul.Error{Reason: "Failed to parse MNR lines", Source: "mnr-parser"}
+		return nil, fmt.Errorf("MNR parser failed to parse any of the %d lines", len(lines))
 	}
 	return metrics, nil
 }
@@ -117,7 +117,7 @@ func (mnr *MNR) mnrParseLine(data []byte) (*skogul.Metric, error) {
 	// First scan should give timestamp or changed symbol
 	ok := scanner.Scan()
 	if !ok {
-		return nil, skogul.Error{Reason: "Failed to extract first value from a MnR line"}
+		return nil, fmt.Errorf("failed to extract first value from a MnR line")
 	}
 	timestamp := scanner.Text()
 
@@ -136,14 +136,14 @@ func (mnr *MNR) mnrParseLine(data []byte) (*skogul.Metric, error) {
 		// The next value should be a timestamp
 		ok := scanner.Scan()
 		if !ok {
-			return nil, skogul.Error{Reason: "Failed to extract timestamp as second value from a MnR line"}
+			return nil, fmt.Errorf("failed to extract timestamp as second value from a MnR line")
 		}
 		timestamp = scanner.Text()
 	}
 
 	tint, err := strconv.ParseInt(timestamp, 10, 64)
 	if err != nil {
-		return nil, skogul.Error{Reason: "Failed to convert string to integer for timestamp", Source: "mnr-parser"}
+		return nil, fmt.Errorf("failed to convert string to integer for timestamp: %w", err)
 	}
 	ts := time.Unix(tint, 0)
 	metric.Time = &ts
@@ -151,7 +151,7 @@ func (mnr *MNR) mnrParseLine(data []byte) (*skogul.Metric, error) {
 	// Fetch MNR group
 	ok = scanner.Scan()
 	if !ok {
-		return nil, skogul.Error{Reason: "Failed to extract MNR group", Source: "mnr-parser"}
+		return nil, fmt.Errorf("failed to extract MNR group")
 	}
 	metric.Metadata["group"] = scanner.Text()
 
@@ -159,7 +159,7 @@ func (mnr *MNR) mnrParseLine(data []byte) (*skogul.Metric, error) {
 	metric.Data = make(map[string]interface{})
 	ok = scanner.Scan()
 	if !ok {
-		return nil, skogul.Error{Reason: "Failed to extract MNR variable name", Source: "mnr-parser"}
+		return nil, fmt.Errorf("failed to extract MNR variable name")
 	}
 	variable := scanner.Text()
 
@@ -169,7 +169,7 @@ func (mnr *MNR) mnrParseLine(data []byte) (*skogul.Metric, error) {
 
 	ok = scanner.Scan()
 	if !ok {
-		return nil, skogul.Error{Reason: "Failed to extract MNR variable value", Source: "mnr-parser"}
+		return nil, fmt.Errorf("failed to extract MNR variable value")
 	}
 	val := parseMNRFieldValue(scanner.Text())
 
--- a/receiver/linefile.go
+++ b/receiver/linefile.go
@@ -65,8 +65,7 @@ func (lf *LineFile) read() error {
 // Start never returns.
 func (lf *LineFile) Start() error {
 	for {
-		err := lf.read()
-		if err != nil {
+		if err := lf.read(); err != nil {
 			lfLog.WithError(err).Error("Unable to read file")
 		}
 		if lf.Delay.Duration != 0 {
--- a/receiver/linefile.go
+++ b/receiver/linefile.go
@@ -47,8 +47,7 @@ type LineFile struct {
 func (lf *LineFile) read() error {
 	f, err := os.Open(lf.File)
 	if err != nil {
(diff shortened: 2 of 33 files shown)


receiver/linefile.go
```diff
@@ -58,16 +57,18 @@ func (lf *LineFile) read() error {
...
 }
 
 // Start never returns.
 func (lf *LineFile) Start() error {
 	for {
-		lf.read()
+		err := lf.read()
+		if err != nil {
+			lfLog.WithError(err).Error("Unable to read file")
+		}
```
sklirg: Good catch.

Not that it really matters here, and it's more of a preference, but it is possible to write this a bit more concise if the data (here: none) and error is not used further on:
```golang
if err := lf.read(); err != nil {
        // body
}
```

receiver/tcpline.go
```diff
@@ -60,11 +59,11 @@ finish up. We should probably add a read-timeout in the future.
 func (tl *TCPLine) Start() error {
 	tcpip, err := net.ResolveTCPAddr("tcp", tl.Address)
 	if err != nil {
-		return skogul.Error{Source: "tcp receiver", Reason: fmt.Sprintf("unable to resolve address %s", tl.Address), Next: err}
```
sklirg, line 63: The default error message by golang here is: "lookup {tl.Address}: Temporary failure in name resolution", e.g. "lookup lolcalhost: Temporary failure in name resolution" (tl.Address = "lolcahost").

I think the current error message is a nice clarification to the user (e.g. in case of a typo), instead of a bit of a more cryptic (IMO) low-level error message.

I think that (example)
`return fmt.Errorf("unable to resolve address %s, err: %w", tl.Address, err)`
could be a nice, user-friendly error message.

receiver/udp.go
```diff
@@ -125,12 +130,10 @@ func (ud *UDP) Start() error {
 
 	udpip, err := net.ResolveUDPAddr("udp", ud.Address)
 	if err != nil {
-		udpLog.WithError(err).WithField("address", ud.Address).Error("Can't resolve address")
```
sklirg, line 128: Same as https://github.com/telenornms/skogul/pull/261/files#r977675191

sender/debug.go
```diff
@@ -54,8 +54,7 @@ func (db *Debug) Send(c *skogul.Container) error {
 	}
 	b, err := e.Encode(c)
 	if err != nil {
-		debugLog.WithError(err).Panic("Unable to encode for debug output")
-		return err
+		return fmt.Errorf("encoding failed: %w", err)
```
sklirg, line 57: If the `Panic()` call here is dropped, this changes how this works (it'll no longer panic). But that's probably fine..? It's also the debug sender, which probably isn't used heavily in regular setups.

sender/fallback.go
```diff
@@ -49,14 +50,18 @@ type Fallback struct {
 }
 
 // Send sends data down stream
+// XXX: Need to log failures?
```
sklirg, line 53: Yes please :D (but not required for this PR)

transformer/timestamp.go
```diff
@@ -70,16 +71,22 @@ func (config *Timestamp) Transform(c *skogul.Container) error {
...
+			// Why log it as ERROR if we don't care enough to
+			// propagate the error up?
 			timestampLogger.Error("Failed to cast timestamp field to a string")
 			if config.Fail {
-				return skogul.Error{Reason: "Failed to cast timestamp field to a string"}
+				return fmt.Errorf("failed to cast timestamp field to a string")
 			}
+			// XXX: Added late, is there a use case where
+			// timestamp isn't a string but we still want to
+			// parse it? O_o
```
sklirg, line 88: I've encountered seconds-since-unix-epoch as a "timestamp" (whether it satisfies as the common standard for "timestamp" or not, I will not dive into - I have seen it!)

See the attached image in https://github.com/telenornms/skogul/issues/242#issue-1324657603 for an example (`.date` field).

sklirg (approved):
LGTM.

Nice job, this is something I remember we talked about a while back, good  to have it finally done.

Most of my comments are nitpicky/questions/comments in general, so they are not blockers for a merge (IMO).

The only concern I have is that a lot of context is lost, but maybe it can be inferred. Such as the custom "Source"-field from `skogul.Error{}`, that the error originated from a sender/handler/parser/receiver. Also some of `logrus.Fields{}` are gone, which could prove useful in debug-cases, but if that's the case I vote for us simply adding a debug/trace log with those required fields when required, instead of adding potential noise to error logs.

parser/mnr.go
```diff
@@ -99,8 +100,7 @@ func (mnr *MNR) mnrParseData(data []byte) ([]*skogul.Metric, error) {
 		metrics = append(metrics, metric)
 	}
 	if len(metrics) == 0 {
-		mnrLog.WithField("lines", len(lines)).Warnf("MNR parser failed to parse any of the %d lines", len(lines))
-		return nil, skogul.Error{Reason: "Failed to parse MNR lines", Source: "mnr-parser"}
+		return nil, fmt.Errorf("MNR parser failed to parse any of the %d lines", len(lines))
```
kamilernerd, line 103: Looks good so far but I noticed this small thing here. Capital letter at the start of an error message.

sender/debug.go
KristianLyng: Good catch... I can't imagine we WANT it to panic, but it certainly changes behavior. I'd say for the better, so leaving this change in.

transformer/timestamp.go
KristianLyng: That's fair enough, but the transformer wouldn't be able to actually parse it as a timestamp, so it makes little sense to try? time.Parse() expects a string, so if the casting fails, it will fail as well.

KristianLyng:
I agree that there are cases where some context might be lost, but in most cases the "context"  just added noise, so I'd rather risk losing a bit of context to gain more overall clarity, and like you said, just fix it when we discover missing context. I did fix the ones you mentioned specifically though.

For logrus, there's a detail I didn't mention: Even when the log-level isn't set to trace, there is still too much work being done just to discard the messages. So while I'd like to use Trace all over the place, it's surprisingly costly. But that's a discussion for an other time I think, and I'd need to investigate further.

I'll give @kamilernerd some time to do a basic review too, just for good measure, but planning on merging it pretty fast so I can start deploying it.

sender/http.go
```diff
@@ -211,7 +207,7 @@ func (ht *HTTP) initStats() {
 // to re-implement them.
 func (ht *HTTP) sendBytes(b []byte) error {
 	if !ht.ok {
-		return skogul.Error{Reason: "HTTP sender not in OK state", Source: "http-sender"}
+		return fmt.Errorf("HTTP sender not in OK state")
```
kamilernerd, line 210: The same as above. Capital letter. Maybe the error message could say something more verbose? Ex. "http sender failed to initialize"

sender/http.go
```diff
@@ -284,14 +277,14 @@ func (ht *HTTP) Send(c *skogul.Container) error {
...
+		return skogul.MissingArgument("URL")
 	}
 	_, err := getCertPool(ht.RootCA)
 	if err != nil {
-		return skogul.Error{Source: "http sender", Reason: fmt.Sprintf("failed to read custom root CA (RootCA: %s)", ht.RootCA), Next: err}
+		return fmt.Errorf("failed to read custom root CA (RootCA: %s): %w", ht.RootCA, err)
 	}
 	if (ht.Certfile != "" && ht.Keyfile == "") || (ht.Certfile == "" && ht.Keyfile != "") {
-		return skogul.Error{Source: "http sender", Reason: "Specify both Certfile and Keyfile if either is specified."}
+		return fmt.Errorf("You must provide BOTH Certfile AND Keyfile, or neither.")
```
kamilernerd: Capital letter

sender/splunk.go
```diff
@@ -177,11 +176,11 @@ func (s *Splunk) Verify() error {
...
 		// missing URL, disregard it, since we will override that
 		// during our own init().
-		if !strings.Contains(err.Error(), "no URL specified") {
-			return skogul.Error{Reason: "Failed to verify HTTP sender for Splunk", Source: "splunk-sender", Next: err}
+		if !strings.Contains(err.Error(), "missing required configuration option `URL'") {
+			return fmt.Errorf("failed to verify HTTP sender for Splunk: %w", err)
 		}
 		if s.HTTP.URL != "" && s.URL != "" && s.HTTP.URL != s.URL {
-			return skogul.Error{Reason: "Splunk URL defined in both HTTP.URL and URL fields with different values. Only specify it once.", Source: "splunk-sender"}
+			return fmt.Errorf("Splunk URL defined in both HTTP.URL and URL fields with different values. Only specify it once.")
```
kamilernerd: Capital letter.

KristianLyng:
Re discussion: I generally agree, however, words that are typically always upper case should remain so, e.g. Splunk is a product name, HTTP and URL are also typically written in upper case, however, "You" should definitely not be upper case for consistency. Fixed.

transformer/timestamp.go
sklirg: Yes, but we could use `time.Unix()` instead for "parsing" this - but we'd need to know if it's in seconds, milliseconds (javascript...), nanoseconds, ... Currently, we don't support this at all.

PR #262, opened 2022-09-23
Feat/sqlite

### LINKED ISSUE #257: Add SQLite driver to sql sender
SSIA

kamilernerd:
In order to test create skogul.sqlite file in testdata directory.


sender/sql.go
```diff
@@ -117,6 +118,12 @@ func (sq *SQL) prep() {
 }
 
 func (sq *SQL) init() {
+	if sq.Driver == "sqlite3" && !SQLiteDBExists(sq.ConnStr) {
+		sq.initErr = fmt.Errorf("failed to initialize. SQLite db might be missing %s", sq.ConnStr)
+		sqlLog.WithError(sq.initErr).WithField("driver", sq.Driver)
```
sklirg: This won't actually log anything, you'll need to call one of the log functions, e.g. `Error(string)`.

sender/sql.go
```diff
@@ -125,6 +132,13 @@ func (sq *SQL) init() {
 	sq.prep()
 }
 
+func SQLiteDBExists(file string) bool {
+	if _, err := os.Stat(file); err != nil {
+		return false
+	}
+	return true
+}
```
sklirg: `os.Stat()` can return an error which can be checked by `os.IsNotExist(err)` for file existence. I think this could be useful here.

Furthermore, I think it can be made private as it is only used by this sender, otherwise, it deserves some comments if it is to be public (denoted by the capital first letter of its name).

Other than that, maybe this could be a more general "connection check" than verifying if a file exists (or not), e.g. being able to open the file etc., and maybe it can be executed in `Verify()` as well.

sklirg (commented):
Nice work. This is sure to be useful in local development where it can be cumbersome to run a full database "just for testing something".

sklirg:
I don't think committing `sender/testdata/skogul.sqlite` is very useful, please add it to a `.gitignore` :)

KristianLyng:
> I don't think committing `sender/testdata/skogul.sqlite` is very useful, please add it to a `.gitignore` :)

Hmm, but it's used in the test cases though, so it either has to be provided or generated on demand for the tests not to fail. I know we don't do test cases for the other SQL drivers normally, but in this case it is mostly trivial... Though it's problematic that it'll get updated with content we don't want to commit...

Maybe copy the original testdata to a tmp-directory/file when running the test cases? Or could it be generated on demand?

KristianLyng (approved):
Looks good to me, not sure if Håkon wants an other stab at it, but I don't see anything missing here.

go.mod
```diff
@@ -13,6 +13,8 @@ require (
 	google.golang.org/protobuf v1.28.1
 )
 
+require github.com/mattn/go-sqlite3 v1.14.15 // indirect
+
```
sklirg: Another nitpick from me; I'm not sure why this is added _here_, and it should be a direct dependency. I'd just add it to the list above and remove the `// indirect` comment (or just run `go mod tidy`, which also will mark it as a direct dependency). I noticed that this happens when using `go get`, so I guess it is fine, just kinda weird. But it should be a direct dependency anyways.

sender/sql.go
```diff
@@ -107,7 +108,7 @@ func (sq *SQL) prep() {
 		} else {
 			sq.list = append(sq.list, dbElement{data, element})
 		}
-		if sq.Driver == "mysql" {
+		if sq.Driver == "mysql" || sq.Driver == "sqlite3" {
```
sklirg, line 111: I think this should accept "sqlite" as well as "sqlite3", as they're pretty much synonyms (as far as I know).

sender/sql.go
```diff
@@ -125,6 +132,21 @@ func (sq *SQL) init() {
...
 
+func verifySQLiteConn(file string) bool {
+	fd, err := os.Stat(file)
+
+	if os.IsNotExist(err) || fd.IsDir() {
+		return false
+	}
+
+	// Check if user has RW permissions to file. Both required.
+	if ((fd.Mode() & 0b010000000) != 0b010000000) || ((fd.Mode() & 0b100000000) != 0b100000000) {
```
sklirg: Consider if this can be made more readable by using e.g. the `io` package, such as [`fs.FileMode`](https://pkg.go.dev/io/fs#FileMode) and some bit-magic.

Furthermore, this checks if _any user_ has rw access to the file (here: root), not necessarily skogul (or the user running skogul), which results in the following:
```
$ stat --format '%a | %A %U:%G %n' /tmp/skogul.sqlite
644 | -rw-r--r-- root:wheel /tmp/skogul.sqlite
$ go run cmd/skogul/main.go cmd/skogul/man.go -f conf.json
ERRO[0000] Failed to transform and send metrics: sending metrics failed: sender failed: attempt to write a readonly database  category=receiver receiver=tester
```
(this should have failed in the init call, not be allowed to proceed on to "send".)
(sender config used:
```json
{
  "type": "sql",
  "driver": "sqlite3",
  "query": "INSERT INTO test VALUES(${timestamp},${metadata.src},${name},${data});",
  "connstr": "/tmp/skogul.sqlite"
}
```
)

Furthermore, on the contrary, it disallows skogul from writing to the database file if the _owner_ of the file does not have read or write access, but skogul does (notice the `rw-` for 'others', but not the owner (root)):
```
$ stat --format '%a | %A %U:%G %n' /tmp/skogul.sqlite
466 | -r--rw-rw- root:wheel /tmp/skogul.sqlite
$ go run cmd/skogul/main.go cmd/skogul/man.go -f conf.json
ERRO[0000]                                               category=sender driver=sqlite3 error="failed to initialize. Verify that /tmp/skogul.sqlite file exists or has correct permissions" sender=sql
ERRO[0000] Failed to transform and send metrics: sending metrics failed: sender failed: database initialization failed: failed to initialize. Verify that /tmp/skogul.sqlite file exists or has correct permissions  category=receiver receiver=tester
```
(I first thought of this as if skogul for some reason did _not_ have `rw-` on the file, but either `group` or `others` _did_, in which case I _think_ skogul still would have access? Which would be an extreme edge case. But then I realised that this applies anyways, because it might not be skogul _owning_ file file, even though it has read/write access - e.g. if it is owned by another system user, but skogul will read/write to it.)

sklirg (changes requested):
Just some bit magic that needs fixing and then this is looking good to me 🪄 😄

sender/sql.go
KristianLyng: Do we even need anything of the verifySQLiteConn function? I mean: Wont the regular connection take care of all relevant checks, because the only real check I care about is "will it work or not"...

go.mod
KristianLyng: Imo the entire go.mod and go.sum thing is a bit of a missfeature the way it works now since its main function is to cause irrelevant merge conflicts in files that are by and large computer generated and maintained, so just run go mod tidy and be done with it I suppose... Maybe we should just do 'go mod tidy' in make check to see if it generates a diff....

sender/sql.go
kamilernerd: I can't agree on this one. There is a big difference between version 2 and 3 of sqlite despite the fact that version 2 is old the use of sqlite as a general name could be a bad choice in case where version 2 would have to be used.

sender/sql.go
KristianLyng: We went through this together and Kamil removed most of the check, but left the basic os.Stat() check and fd.IsDir, this is mainly to prevent panics, but everything else is caught by db.Open() now, and seems to report correctly. One last look, @sklirg ?

sender/sql.go
sklirg: I hope nobody runs sqlite2, its last release seems to be from 2005. But the configuration check is very helpful and will guide the user to the correct term, so I am fine with keeping it as-is.

Example error:
```
FATA[0000] Failed to configure Skogul                    category=cmd cmd=main error="configuration for sender `sqlite' doesn't verify: unsuported database driver sqlite - must be `mysql` or `postgres` or `sqlite3`"
```

sender/sql.go
sklirg: LGTM 🎉

sklirg (approved):
LGTM :shipit:

PR #263, opened 2022-09-27
Add Prometheus parser

KristianLyng:
Work mainly by @RoshiniNara 

This parses a Prometheus-document into individual metrics per line/variable. There's some discussion on if we could re-use a metric with shared data, but that seems like something slightly out side the scope of the basic parser for now, so I'm leaving that as a separate task, but feedback on whether this would actually be usable would be nice.


KristianLyng:
Ok, there are intermittent failures here, I'm investigating.

KristianLyng:
So the tests are failing because three fields results in three different metrics, and the order will vary.

This is clearly a test-issue, but it is, somewhat less clearly, a bit of a challenge in the parser as well.

Data in a prometheus data share metadata, but each field/metric can add additional metadata. It is thus technically correct to distinguish each field/stats in a separate metric, but it is not necessarily the best option.

De-duplication could be an option, but adds a non-trivial amount of complexity to the parser and non-obvious memory pressure. 

I think the correct solution here is to accept the parser (with fixed tests) as is, and rather add a de-duplication transformer/mechanism.

KristianLyng:
@sklirg I've reviewed this and while there are style-changes here I'd like to see, I think the core functionality works. Since I seem to remember you talking about this, I'd like some input on the usability when each variable ends up as a separate metric though... I have an idea for a general-purpose "joiner" transformer that might be able to do the trick though...

parser/prometheus.go
```diff
@@ -0,0 +1,85 @@
...
+			if err != nil {
+				return nil, err
+			}
+			// we do not need tmp to iteratate to get to the value. The library offers GetUntyped().Value call to directly get the value.
+			dataDict[k] = i.GetUntyped().Value
+			dataDictstr, _ := json.Marshal(dataDict)
+			err1 := json.Unmarshal(dataDictstr, &tmpMetric[indexCounter].Data)
+			if err1 != nil {
+				return nil, err
+			}
```
sklirg, line 76: `err` can be re-assigned (`=`) here (instead of (re-)initialized (`:=`), which I assume was tried before, as that will make the compiler complain that `err` is already initialised). The previous `err` is either `nil` or has been returned already by the previous `if err != nil` branch.

Re-using the `err` variable is cleaner than to add a number to the variable name. Furthermore, this branch will return a `nil` err, as it returns `err` and not `err1`.

parser/prometheus.go
```diff
@@ -0,0 +1,85 @@
...
+			container.Metrics = append(container.Metrics, &tmpMetric[indexCounter])
+			// clean up the old values of the dictionary so that they don't get carried to the next iteration.
+			metadataDict = make(map[string]interface{})
+			dataDict = make(map[string]interface{})
```
sklirg, line 80: Can this be initialised in the start of the loop instead of initialised before the loops and then cleaned up at the end of each one?

parser/prometheus_test.go
```diff
@@ -0,0 +1,129 @@
...
+	p := parser.Prometheus{}
+
+	container, err := p.Parse(b)
+
+	if err != nil {
+		t.Logf("Error while parsing the data. %v", err)
+		t.FailNow()
+	}
+
+	//expectedTime := 1608520832877
```
sklirg, line 51: Commented out code can be removed. Otherwise, it deserves a comment about why it is useful to keep around (e.g. if it is used for debugging some test case in some cases).

parser/prometheus_test.go
```diff
@@ -0,0 +1,129 @@
...
+		if m.Data["net_conntrack_dialer_conn_attempted_total"] != nil {
+			m1 = m
+		}
+		if m.Data["conntrack_dialer_conn_attempted_total"] != nil {
+			m2 = m
+		}
+		if m.Data["dialer_conn_attempted_total"] != nil {
+			m3 = m
+		}
+	}
```
sklirg, line 87: Considering that we know that the container contains 3 metrics already (based on the previous checks), we can assign this directly.

E.g. `m1 := container.Metrics[0].Data["net_conntrack_dialer_conn_attempted_total"]`

parser/prometheus_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	if m1 == nil || m2 == nil || m3 == nil {
+		t.Errorf("missing metrics?")
```
sklirg, line 90: This comment goes for the following test cases as well.

It's nice to explain what was expected and exactly what failed, so that when/if these tests fail some time in the future, it's easy to understand what was expected and what was received. The test cases for `mN.Data[dataNKey]` are well created, so if these for general metrics & metrics metadata could be updated to look like them, that'd be great.

Issue #182, opened 2021-10-12
Rename the json parser


sklirg:
The parser you get when you specify `parser: json` in config today is a parser of the opinionated skogul json format[0].

I think we should rename this parser to "skogul", and we can rename the "custom json" format to "json". That way, "generic" JSON data can be parsed using the "json" parser, and the opinionated skogul format can be parsed by using "skogul".

This should be done in multiple steps to help with the breaking change:

1) Rename the json parser to "skogul".
  *) Add an alias "json" for the current skogul parser.
  *) Add a deprecation warning for using the now-outdated json parser (i.e. using the json parser)
2) Allow people time to switch to the new "skogul" parser
3) Remove the "json"-alias, and rename the "custom-json" parser to "json".

0:
```json
{ "timestamp": "2021-01-01T00:00:00Z"
, "metadata": {}
, "data": {}
}

PR #267, opened 2022-10-26
chore: Deprecate 'json' alias for Skogul json parser

sklirg:
This change renames the currently existing JSON parsers.

The 'skogul JSON' parser is renamed (in code) to 'SkogulJSON', and in configuration it only accepts
"skogul" (instead of "skogul" and "json"). Using "json" still works, but produces a warning.

Following this, to avoid confusion in the code (mainly for new contributors), the existing 'RawJSON' struct
is renamed to 'JSON'. This might cause confusion for existing contributors used to 'JSON' being 'Skogul JSON',
but hopefully this won't be too bad. I did this change in a separate commit, so it is easy to drop this if we
want to keep the 'RawJSON' name (or if we want to change it to something else).

This is a start on the road of #182. I suggest a release or two before continuing. :-)

### DIFF
--- a/common_test.go
+++ b/common_test.go
@@ -193,7 +193,7 @@ func TestParseInvalidContainerAndTransformItValid(t *testing.T) {
 	data := []byte(fmt.Sprintf(`{"data": 1, "ts": "%s"}`, timestring))
 
 	h := skogul.Handler{}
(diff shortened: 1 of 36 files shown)


KristianLyng:
So I've reviewed this more than once and I see no fault here, and I agree with it. Merging.

PR #270, opened 2022-10-31
Feat/unflatten

kamilernerd:


KristianLyng:
Care to run "make fmtfix" to solve the formating issues?

transformer/auto.go
```diff
@@ -108,4 +108,11 @@ func init() {
 		Help:     "PROTOTYPE/ALPHA: Static enrichment. Use a json-structured source document to enrich incoming metrics with additional data. See the docs/examples/ directory for an actual example. This is NOT production ready, and should only be used if you are prepared to file bug reports and update your setup as the transformer matures.",
 		AutoMake: true,
 	})
+	Auto.Add(skogul.Module{
+		Name:     "unflatten",
+		Aliases:  []string{},
+		Alloc:    func() interface{} { return &Unflatten{} },
+		Help:     "Create structured data from flat structure",
```
KristianLyng: I think some slightly more detailed explanation could be useful here. E.g.: "Create structured data from flat keys, e.g.: foo.bar.zoo: 1, foo.baz.zoo: 2 to { foo: { bar: { zoo: 1 }, baz: { zoo: 2 } } }" ?

transformer/unflatten.go
```diff
@@ -0,0 +1,64 @@
...
+
+// Transform created a container of metrics
+func (u *Unflatten) Transform(c *skogul.Container) error {
+	for mi := range c.Metrics {
+		c.Metrics[mi] = u.convertValues(c.Metrics[mi])
+	}
+
+	return nil
+}
+
```
KristianLyng, line 24: Given the complexity of coverteValues, a human-readable description of the algorithm would be helpful here for future debugging.

transformer/unflatten.go
```diff
@@ -0,0 +1,64 @@
...
+	return newMetric
+}
+
+func (u *Unflatten) recursivelyCreateMap(root map[string]interface{}, keys []string, value interface{}, pos int) interface{} {
+	if pos == len(keys) {
+		return value
+	}
+
+	if pos < len(keys) {
```
KristianLyng: Minor nitpick: the go-ism would be to do if pos > len(keys) { return root }, followed by the code, now without indentation. No functional change, just less indentation.

In this case, that shouldn't even be possible, so the whole if() should be pointless, though I suppose it doesn't hurt, but should probably be a log/error if it fails since it should be impossible to arrive here with pos > len(keys).

transformer/unflatten_test.go
```diff
@@ -0,0 +1,50 @@
...
+	log "github.com/sirupsen/logrus"
+	"github.com/telenornms/skogul/parser"
+	"github.com/telenornms/skogul/transformer"
+	"os"
+	"testing"
+)
+
+func readFile(file string, t *testing.T) []byte {
+	t.Helper()
+	b := make([]byte, 9000)
```
KristianLyng: You can just use os.ReadFile() and not bother with most of this.

transformer/unflatten_test.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	u := transformer.Unflatten{}
+	err = u.Transform(container)
+
+	jsonData, err := json.MarshalIndent(container.Metrics[0].Data["Report"], "", "	")
+
+	if err != nil {
+		t.Error("Could not json marshal container")
+	}
```
KristianLyng: This test case just checks that the transformer runs and doesn't crash, maybe check if a single value is present at least?

transformer/unflatten_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	log.Println(string(jsonData))
```
KristianLyng: Please use t.Log() instead, it'll handle output better.

KristianLyng (changes requested):
I think the implementation looks good, well done!

While I made several comments the important ones for me to accept the PR is the lack of comments, since this is a bit of an intricate problem, despite what ended up being a nice and elegant solution.

I'd also like to se at least a single verification in the test case of actual data stored where it's supposed to be.

Other than that, the other comments are nice-to-have, not need-to-have.

sklirg:
This is more of a general question than a review: Is there a reason for using `.` (dots) for splitting the data here, and not e.g. `/` (forward slash) which is what the [JSON Pointer RFC 6901](https://datatracker.ietf.org/doc/html/rfc6901) uses or allowing users to customize this in general?

I think it would be beneficial to support RFC 6901 as well, either as the baseline or as an alternative, e.g. by allowing users to configure the "path separation character" (RFC 6901 adds in some more complexity by adding escape chars etc).

Example usage: If this is to be used with the `data.Flatten` transformer, the default separator for that is `__`, and this implementation uses `.`, so using flatten and then unflatten would not work unless flatten was explicitly configured to use `.` as the flatten separator.

KristianLyng:
@sklirg First, I agree that we should have matching fetaures for flatten and unflatten.

Second: This is based on actual source data which uses dots, which is the origin for this. This isn't an argument for anything, just an explanation of where we're coming from.

Thirdly: The json-pointers is, if I understand it correctly, to refer to content inside a json tree, while this is... not quite that.

However.

@kamilernerd I suggest you add a configurable for the separator, so we can (theoretically, at least) combine the flatten and unflatten transformer.

PR #271, opened 2022-10-31
Feat/ip tool

kamilernerd:


KristianLyng (changes requested):
So the checks aren't running, that needs to be resolved.

Additionally, I'd like us to support both string-encoded IP-addresses, and net.IP: you're already pretty close, so shouldn't be a problem.

One bigger problem: You return int() from inet6Aton(): That's either 32-bit or 64-bit, depending on architecture. IPv6 uses 128-bit addresses. (I'm pretty sure ipInt.IsInt64() would reveal this). I think just returning math/big.Int is acceptable: It implements json, text and gob marshaling/unmarshaling, which should be good enough.

PR #272, opened 2022-11-07
Feat/ban values

kamilernerd:


KristianLyng (changes requested):
We did a on-site review and the short version is that there are some changes that are needed.

What we want to achieve is:

```
{
    "foo": "bar"
    "enabled": true,
     "bytes": 123125425,
}
```
And ban the ENTIRE metric based on enabled true/false, while what this PR does is just ban the actual field, which is not quite what we want.

A new PR will be prepared :D

PR #275, opened 2022-11-18
Feat/nats pubsub

n-holmstedt:
Telenor Sweden is looking to use skogul in a number of usecases, hence this PR with basic NATS pub/sub (at least one delivery). 

* sender/nats.go - Basic sender with rudimentary config test
* receiver/nats.go - Basic receiver with rudimentary config test
* I introduced the "GetCertPool" function (originally found in http-module) in the common module for usage across modules.

Also: The number of hours i've spent on golang is less than i have fingers on one hand, so do not hesitate to give me pointers. Most of the code is pretty much copy/paste from your own modules (which was really easy to follow).

### DIFF
--- a/common.go
+++ b/common.go
@@ -384,7 +384,7 @@ func MissingArgument(field string) error {
 	return fmt.Errorf("missing required configuration option `%s'", field)
 }
 
-// Make GetCertPool a common function.
+// GetCertPool accepts a path to a directory of certificate authorities to trust. Pass in the empty string to use system defaults
 func GetCertPool(path string) (*x509.CertPool, error) {
 	// this means "use system default"
 	if path == "" {
--- a/common.go
+++ b/common.go
@@ -25,11 +25,11 @@
 package skogul
 
 import (
-	"os"
+	"crypto/x509"
 	"fmt"
 	"math"
+	"os"
 	"runtime"
-        "crypto/x509"
 
 	"github.com/sirupsen/logrus"
 )
@@ -384,29 +384,28 @@ func MissingArgument(field string) error {
 	return fmt.Errorf("missing required configuration option `%s'", field)
 }
 
-//Make GetCertPool a common function.
+// Make GetCertPool a common function.
 func GetCertPool(path string) (*x509.CertPool, error) {
-        // this means "use system default"
-        if path == "" {
-                return nil, nil
-        }
-        cp := x509.NewCertPool()
-        fd, err := os.Open(path)
-        if err != nil {
-                return nil, fmt.Errorf("unable to open custom root CA: %w", err)
-        }
-        defer func() {
-                fd.Close()
-        }()
-        bytes := make([]byte, 1024000)
-        n, err := fd.Read(bytes)
-        if err != nil {
-                return nil, fmt.Errorf("unable to read custom root CA: %w", err)
-        }
-        ok := cp.AppendCertsFromPEM(bytes[:n])
-        if !ok {
-                return nil, fmt.Errorf("unable to append certificate to root CA pool")
-        }
-        return cp, nil
+	// this means "use system default"
+	if path == "" {
+		return nil, nil
+	}
+	cp := x509.NewCertPool()
+	fd, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("unable to open custom root CA: %w", err)
+	}
+	defer func() {
+		fd.Close()
+	}()
+	bytes := make([]byte, 1024000)
+	n, err := fd.Read(bytes)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read custom root CA: %w", err)
+	}
+	ok := cp.AppendCertsFromPEM(bytes[:n])
+	if !ok {
+		return nil, fmt.Errorf("unable to append certificate to root CA pool")
+	}
+	return cp, nil
 }
-
--- a/common.go
+++ b/common.go
@@ -384,7 +384,7 @@ func MissingArgument(field string) error {
 	return fmt.Errorf("missing required configuration option `%s'", field)
 }
 
-//Make GetCertPool a common function instead of reciever specific.
+//Make GetCertPool a common function.
 func GetCertPool(path string) (*x509.CertPool, error) {
         // this means "use system default"
         if path == "" {
--- a/common.go
+++ b/common.go
@@ -25,9 +25,11 @@
 package skogul
 
 import (
+	"os"
 	"fmt"
 	"math"
 	"runtime"
+        "crypto/x509"
 
 	"github.com/sirupsen/logrus"
 )
@@ -381,3 +383,30 @@ func IsInf(f float32, sign int) bool {
 func MissingArgument(field string) error {
 	return fmt.Errorf("missing required configuration option `%s'", field)
 }
+
+//Make GetCertPool a common function instead of reciever specific.
+func GetCertPool(path string) (*x509.CertPool, error) {
+        // this means "use system default"
+        if path == "" {
+                return nil, nil
+        }
+        cp := x509.NewCertPool()
+        fd, err := os.Open(path)
+        if err != nil {
+                return nil, fmt.Errorf("unable to open custom root CA: %w", err)
+        }
+        defer func() {
+                fd.Close()
+        }()
+        bytes := make([]byte, 1024000)
+        n, err := fd.Read(bytes)
+        if err != nil {
+                return nil, fmt.Errorf("unable to read custom root CA: %w", err)
+        }
+        ok := cp.AppendCertsFromPEM(bytes[:n])
+        if !ok {
+                return nil, fmt.Errorf("unable to append certificate to root CA pool")
+        }
+        return cp, nil
+}
+
--- a/docs/examples/payloads/nats-payload.json
+++ b/docs/examples/payloads/nats-payload.json
@@ -1,0 +1,1 @@
+{"metrics":[{"timestamp":"{{TimeStamp}}","metadata":{"id":"{{ID}}"},"data":{"string":"{{Random 5 10}}","float":1.11,"integer":{{Count}}}}]}
--- a/docs/examples/payloads/nats-payload.json
+++ b/docs/examples/payloads/nats-payload.json
@@ -1,1 +1,0 @@
-{"metrics":[{"timestamp":"{{TimeStamp}}","metadata":{"id":"{{ID}}"},"data":{"string":"{{Random 5 10}}","float":1.11,"integer":{{Count}}}}]}
--- a/receiver/nats.go
+++ b/receiver/nats.go
@@ -23,12 +23,9 @@
 package receiver
 
 import (
-	"fmt"
-	"os"
 	"github.com/nats-io/nats.go"
 	"github.com/telenornms/skogul"
 	"crypto/tls"
-	"crypto/x509"
 )
 
 var natsLog = skogul.Logger("receiver", "nats")
@@ -70,32 +67,6 @@ func (n *Nats) Verify() error {
 	return nil
 }
 
-//Unsure on how "Self contained" the receiver should be.
-func getCertPool(path string) (*x509.CertPool, error) {
-	// this means "use system default"
-	if path == "" {
-		return nil, nil
-	}
-	cp := x509.NewCertPool()
-	fd, err := os.Open(path)
-	if err != nil {
-		return nil, fmt.Errorf("unable to open custom root CA: %w", err)
-	}
-	defer func() {
-		fd.Close()
-	}()
-	bytes := make([]byte, 1024000)
-	n, err := fd.Read(bytes)
-	if err != nil {
-		return nil, fmt.Errorf("unable to read custom root CA: %w", err)
-	}
-	ok := cp.AppendCertsFromPEM(bytes[:n])
-	if !ok {
-		return nil, fmt.Errorf("unable to append certificate to root CA pool")
-	}
-	return cp, nil
-}
-
 func (n *Nats) Start() error {
 	if n.Name == "" {
 		n.Name = "skogul"
@@ -130,7 +101,7 @@ func (n *Nats) Start() error {
 			return err
 		}
 
-		cp, err := getCertPool(n.TLSCACert)
+		cp, err := skogul.GetCertPool(n.TLSCACert)
                 if err != nil {
                         natsLog.Fatalf("Failed to initialize root CA pool")
 			return err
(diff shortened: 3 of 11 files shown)


receiver/nats.go
```diff
@@ -0,0 +1,164 @@
...
+		n.Name = "skogul"
+	}
+	n.o = &[]nats.Option{nats.Name(n.Name)}
+
+	if n.Servers == "" {
+		n.Servers = nats.DefaultURL
+	}
+
+	//User Credentials
+	if n.UserCreds != "" && n.NKeyFile != "" {
```
KristianLyng: This could be part of Verify as well.

receiver/nats.go
```diff
@@ -0,0 +1,164 @@
...
+			natsLog.Warnf("Using plain text password over a non encrypted transport!")
+		}
+		*n.o = append(*n.o, nats.UserInfo(n.Username, n.Password))
+	}
+	//TLS authentication, Note: Fix selfsigned certificates.
+	if n.TLSClientKey != "" && n.TLSClientCert != "" {
+		cert, err := tls.LoadX509KeyPair(n.TLSClientCert, n.TLSClientKey)
+		if err != nil {
+			natsLog.Fatalf("error parsing X509 certificate/key pair: %v", err)
+			return err
```
KristianLyng: Since it's returning error, it's usually better to just do return fmt.Errorf("error parsing ..."), otherwise it will cause duplicate error messages on the CLI. (which we've been trying to clean up for a while now :D)

receiver/nats.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+
+		cp, err := skogul.GetCertPool(n.TLSCACert)
+                if err != nil {
+                        natsLog.Fatalf("Failed to initialize root CA pool")
+			return err
```
KristianLyng: Same as above with duplicate error messages.

sender/nats.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	n.o = &[]nats.Option{nats.Name(n.Name)}
+
+	if n.Servers == "" {
+		n.Servers = nats.DefaultURL
+	}
+
+	//User Credentials, use either.
+	if n.UserCreds != "" && n.NKeyFile != "" {
+		natsLog.Fatal("Please configure usercreds or nkeyfile.")
```
KristianLyng: Should be part of Verify so it fails on skogul -show, not only on attempted start.

sender/nats.go
```diff
@@ -0,0 +1,171 @@
...
+			natsLog.Warnf("Using plain text password over a non encrypted transport!")
+		}
+		*n.o = append(*n.o, nats.UserInfo(n.Username, n.Password))
+	}
+
+	//TLS authentication, Note: Fix selfsigned certificates.
+	if n.TLSClientKey != "" && n.TLSClientCert != "" {
+		cert, err := tls.LoadX509KeyPair(n.TLSClientCert, n.TLSClientKey)
+		if err != nil {
+			natsLog.Fatalf("error parsing X509 certificate/key pair: %v", err)
```
KristianLyng: Fatal/killing on failure is not ideal. For setups with multiple pipelines, it would cause serious issues. It could mean that a skogul instance with 3 unrelated receiver-sender pipelines fail to start because just one of them fail.

If possible, I suggest storing the error on n.init_error = err, then in Send() checking if n.init_error exists and return that. It will be slightly less than ideal for setups with just a single receiver and sender, but the log messages should still be obvious.

sender/nats.go
```diff
@@ -0,0 +1,171 @@
...
+		n.init()
+	})
+	nm := make([]byte, 0, len(c.Metrics))
+	for _, m := range c.Metrics {
+		b, err := n.Encoder.E.EncodeMetric(m)
+		if err != nil {
+			return fmt.Errorf("couldn't encode metric: %w", err)
+		}
+		nm = append(nm, b...)
+	}
```
KristianLyng, line 183: I think you want to use just regular n.Encoder.E.Encode(c) for this. We use EncodeMetric(m) in kafka because Kafka has implicit batching, but here you are just concatinating each metric after the other, which means there's no clear separation between them and will probably make the parsing somewhat problematic.

KristianLyng (changes requested):
I'm fairly happy with this, there are some relatively minor comments but the overall design looks solid, though I have been unable to test it myself.

A couple of comments though:

1. Make sure to run 'make fmtfix' if you haven't already to fix any formating issues
2. Unless I'm misreading something, you want to use E.Encode() instead of E.EncodeMetric()

n-holmstedt:
Think i've managed your points here, and i also added the function to append subjects with configuration-defined metadata-fields.

common.go
```diff
@@ -381,3 +383,29 @@ func IsInf(f float32, sign int) bool {
 func MissingArgument(field string) error {
 	return fmt.Errorf("missing required configuration option `%s'", field)
 }
+
+// Make GetCertPool a common function.
+func GetCertPool(path string) (*x509.CertPool, error) {
```
sklirg: Go automatically documents public functions (Capitalized function name) when the comment above it starts with the same name. Please change the comment to start with "// GetCertPool [...]", e.g. `// GetCertPool accepts a path to a directory of certificate authorities to trust. Pass in the empty string to use system defaults`

docs/examples/payloads/nats-payload.json
```diff
@@ -0,0 +1 @@
+{"metrics":[{"timestamp":"{{TimeStamp}}","metadata":{"id":"{{ID}}"},"data":{"string":"{{Random 5 10}}","float":1.11,"integer":{{Count}}}}]}
```
sklirg: Nitpick: I don't know nats, but this seems to be invalid JSON. If it is not JSON, change the file name of the file to a more appropriate type.

receiver/nats.go
```diff
@@ -0,0 +1,163 @@
...
+var natsLog = skogul.Logger("receiver", "nats")
+
+/*
+Nats.io simple pub/sub receiver with:
+
+- Authentication: Username/Password, TLS
+- Authorization: Username/Password, UserCredentials/JWT
+- Queue: Load balancing for multiple receivers in the same Queue.
+*/
+type Nats struct {
```
sklirg: Same comment regarding Golang Docs as above.

Additionally, I'm not sure what the bullet list refers to? I can't see "Authentication" or "Authorization" referenced in the configuration by those names. Could you update the documentation with which does what?

receiver/nats.go
```diff
@@ -0,0 +1,163 @@
...
+	Handler       skogul.HandlerRef `doc:"Handler used to parse, transform and send data."`
+	Servers       string            `doc:"Comma separated list of nats URLs"`
```
sklirg, line 41: Use a `[]string` as a list of strings instead of manually comma-separating this to be in line with other Skogul configuration.

If this is how nats accept the configuration, and it is common to use this format, I can accept keeping it this way. Otherwise, I'm more inclined to use `strings.Join()` or something similar to join them back together.

receiver/nats.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (n *Nats) Start() error {
+	if n.Name == "" {
+		n.Name = "skogul"
+	}
+	n.o = &[]nats.Option{nats.Name(n.Name)}
+
+	if n.Servers == "" {
+		n.Servers = nats.DefaultURL
+	}
```
sklirg: I'm not sure what the default URL of nats is, and if it is common to omit the URL used for connecting.

In general I'm more for being explicit, so I'd suggest making "Servers" a required configuration option (by adding a `Verify()` check which fails if `n.Servers` is not configured.)

receiver/nats.go
```diff
@@ -0,0 +1,163 @@
...
+	Name          string            `doc:"Client name"`
+	Username      string            `doc:"Client username"`
+	Password      string            `doc:"Client password"`
+	TLSClientKey  string            `doc:"TLS client key file path"`
+	TLSClientCert string            `doc:"TLS client cert file path"`
+	TLSCACert     string            `doc:"CA cert file path"`
+	UserCreds     string            `doc:"Nats credentials file path"`
+	NKeyFile      string            `doc:"Nats nkey file path"`
+	Insecure      bool              `doc:"TLS InsecureSkipVerify"`
+	o             *[]nats.Option
```
sklirg: Please use a more descriptive variable name for the struct key.

receiver/nats.go
```diff
@@ -0,0 +1,163 @@
...
+	nc            *nats.Conn
```
sklirg: Please use a more descriptive variable name for the struct key.

sklirg (changes requested):
I did a first-pass on this with some general feedback. Feel free to ask if you have any questions!

receiver/nats.go
n-holmstedt: Authorization and Authentication (AuthZ, AuthN) is separated within Nats, so it's more to show that these are implemented. Buy i guess i can document it as a single, since it isn't explicitly referenced in the code.

receiver/nats.go
n-holmstedt: Yes, exactly. This is the way all the clients behave, as well as the official cli binaries. Otherwise i probably would have separated them.

receiver/nats.go
n-holmstedt: Same as above. It's pretty much how they do it everywere. I don't really have any issue with changing it tho.

It defaults to localhost.

docs/examples/payloads/nats-payload.json
n-holmstedt: Technically correct is the best kind of correct.

receiver/nats.go
```diff
@@ -0,0 +1,159 @@
...
+		if err := n.Handler.H.Handle(msg.Data); err != nil {
+			natsLog.WithError(err).Warn("Unable to handle Nats message")
+		}
+		return
+	}
+
+	n.wg.Add(1)
+	if err != nil {
+		natsLog.Errorf("Encountered an error while connecting to Nats: %v", err)
+	}
```
sklirg: I think this err-check should live closer to the place it was generated; right after `nats.Connect(...)`. The callback function in-between here can be defined elsewhere so that it is "not in the way" (e.g. above the connect).

sender/nats.go
```diff
@@ -0,0 +1,181 @@
...
+
+	//Plain text passwords
+	if n.Username != "" && n.Password != "" {
+		if n.TLSClientKey != "" {
+			natsLog.Warnf("Using plain text password over a non encrypted transport!")
+		}
+		*n.conOpts = append(*n.conOpts, nats.UserInfo(n.Username, n.Password))
+	}
+
+	//TLS authentication, Note: Fix selfsigned certificates.
```
sklirg: Is this "Note:" fixed?

receiver/nats.go
```diff
@@ -0,0 +1,159 @@
...
+		if err != nil {
+			natsLog.Fatal(err)
+		}
+		*n.conOpts = append(*n.conOpts, opt)
+	}
+
+	//Log disconnects
+	*n.conOpts = append(*n.conOpts, nats.DisconnectErrHandler(func(nc *nats.Conn, err error) {
+		natsLog.WithError(err).Error("Got disconnected!")
+	}))
```
sklirg, line 127: Is it always an error if a client disconnects? I'm thinking of controlled restarts of the client or similar, e.g. maintenance/patching. Or will that not trigger a "disconnect" but rather a "bye" of some sort?

I'm thinking that if this logs an error when a client "cleanly disconnects", it's noisy. But if this doesn't trigger on a "clean disconnect", I'm fine with it.

sender/nats.go
```diff
@@ -0,0 +1,181 @@
...
+	}))
+	//Always try reconnecting
+	*n.conOpts = append(*n.conOpts, nats.RetryOnFailedConnect(true))
+	//Keep doing reconnects
+	*n.conOpts = append(*n.conOpts, nats.MaxReconnects(-1))
+
+	var err error
+	n.natsCon, err = nats.Connect(n.Servers, *n.conOpts...)
+	if err != nil {
+		n.init_error = fmt.Errorf("Encountered an error while connecting to Nats: %v", err)
```
sklirg, line 149: This formatting directive could be `%w` (not `%v`) to enable Golang error wrapping. (See the section [here](https://go.dev/blog/go1.13-errors) about Wrapping for details, and the follow-up for when to do it/not to do it.)

sender/nats.go
```diff
@@ -0,0 +1,181 @@
...
+		if n.SubjectAppend != "" && a_subject.(string) != "" {
+			if !strings.HasSuffix(m.Metadata[n.SubjectAppend].(string), ".") {
+				subject = subject + "."
+			}
+			subject = subject + m.Metadata[n.SubjectAppend].(string)
+		}
+
+		b, err := n.Encoder.E.EncodeMetric(m)
+		if err != nil {
+			natsLog.Warnf("couldn't encode metric: %w", err)
```
sklirg: Please use `.WithErr(err).Warn(...)` here instead. And consider if more detail can be added, such as some metadata, the ID in the list of Metrics, or similar, to make it easier to debug if this happens (this can be logged on another log-level so that it is possible to enable this conditionally while debugging).

sklirg (changes requested):
I did a new pass of everything now, some smaller things but other than that this looks good to me 👍

receiver/nats.go
n-holmstedt: Good point, i had not verified it, but the API has an "DisconnectHandler" which seem to be invoked on every disconnect, rather than if an error occurred as in the case of DisconnectErrHandler. 

Did a quick local test as well, which confirmed that it wont trigger on a "clean" disconnect.

sender/nats.go
n-holmstedt: Good point!

sender/nats.go
n-holmstedt: Yea, lingering comments since the first tests with TLS. Fixed with the "Insecure(SkipVerify)" in the configuration.

sklirg (approved):
Code LGTM, but I haven't tested the implementation since I don't know Nats 🥷

sender/nats.go
```diff
@@ -0,0 +1,186 @@
...
+					continue
+				}
+				if !strings.HasPrefix(appSubject.(string), ".") {
+					appSubject = "." + appSubject.(string)
+				}
+				subject = subject + appSubject.(string)
+			}
+		}
+
+		b, err := n.Encoder.E.EncodeMetric(m)
```
KristianLyng, line 176: This needs to be n.Encoder.E.Encode() instead.

KristianLyng (changes requested):
I am happy with this PR now, EXCEPT that the sender needs to use e.Encode() instead of EncodeMetric.

EncodeMetric is mostly reserved for protocols that provide internal batching that superseedes skogul's batching. By using it instead of Encode(), you are effectively disabling batching.


.... 

Edit: Never mind, I now see that this is all about SubjectAppend that we talked about previously.

PR #276, opened 2022-11-24
Replace old drone config with gh actions

kamilernerd:


sklirg:
Do you have any example runs of this action anywhere? If you have not, I suggest forking Skogul and trying to get some of the jobs to run e.g. by pushing tags or by changing the run condition.

In general, the steps look fine to me. But without seeing an example run, it's just config to me ;) I suggest removing the drone config as well :)

On a higher level, I'm hesitant about using actions from "random people" (as in: not from GitHub itself or from more prominent organizations, as it adds more maintenance overhead in debugging if (when...) they break, figuring out when to upgrade, following security notices, etc.), but considering that the [official one from GitHub is archived, and they suggest some actions from "random people"](https://github.com/actions/create-release), I ... don't really care. 😬

kamilernerd:
> Do you have any example runs of this action anywhere? If you have not, I suggest forking Skogul and trying to get some of the jobs to run e.g. by pushing tags or by changing the run condition.
> 
> In general, the steps look fine to me. But without seeing an example run, it's just config to me ;) I suggest removing the drone config as well :)
> 
> On a higher level, I'm hesitant about using actions from "random people" (as in: not from GitHub itself or from more prominent organizations, as it adds more maintenance overhead in debugging if (when...) they break, figuring out when to upgrade, following security notices, etc.), but considering that the [official one from GitHub is archived, and they suggest some actions from "random people"](https://github.com/actions/create-release), I ... don't really care. 😬

I do have tests/results of this in a private repository. Can provide if requested.
Agree, we could make our own should not be a lot of work. I just wanted to get this out quick.

sklirg (approved):
If the deployment actually works, this looks good to me 👍 :shipit:

PR #279, opened 2022-11-25
fix: Allow GitHub Actions ample time to run specific tests

sklirg:
See example usage in 0a2aec73b117ac63572c62fba0e2b1a57a5fcf07 and in the workflow updated


sklirg:
I can update the other timing-sensitive tests with this as well if you wish. Luckily, they have more timings in the test errors so it's easier to figure out approximate timings that should work :)

PR #286, opened 2023-01-27
Allow reading bucket from metadata

kamilernerd:


KristianLyng:
We've now gone through this a few times and because this means basing the URL on a single metric, this wont really work as expected. We've instead decided to go for the approach of hard-coding buckets for now, and using the switch-sender to pick one. Feel free to open an issue if this is unclear or other comments are needed.

PR #289, opened 2023-02-23
Rabbitmq queue

kamilernerd:


sender/rabbitmq_test.go
```diff
@@ -0,0 +1,86 @@
...
+	metrics := make([]*skogul.Metric, 0)
+	metrics = append(metrics, &metric)
+
+	return &skogul.Container{
+		Metrics: metrics,
+	}
+}
+
+func TestRabbitmq(t *testing.T) {
+	data := createContainer()
```
sklirg: You can skip this test in CI because there is no rabbitmq-listener to send to using `if testing.Short()` like this: https://github.com/telenornms/skogul/blob/primary/sender/sql_test.go#L147

That'll make the CI checks ✅

receiver/rabbitmq.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"fmt"
+
+	amqp "github.com/rabbitmq/amqp091-go"
+	"github.com/telenornms/skogul"
+)
+
+type Rabbitmq struct {
+	Username string             `doc:"Username for rabbitmq instance"`
+	Password string             `doc:"Password for rabbitmq instance"`
```
sklirg: This could use the `Secret` type (instead of string) to avoid being (accidentally) exposed in logs etc: https://github.com/telenornms/skogul/blob/primary/common.go#L348

receiver/rabbitmq.go
```diff
@@ -0,0 +1,111 @@
...
+		r.Port = "5672"
+	}
+
+	if r.Host == "" {
+		r.Host = "localhost"
+	}
+
+	if r.Handler == nil {
+		r.Handler = &skogul.HandlerRef{}
+	}
```
sklirg: If no handler is configured, does it make sense to create an empty handler? This makes Skogul receive the messages, but not do anything with them.

I think it could make more sense to implement Verify and check that a handler is set.

receiver/rabbitmq_test.go
```diff
@@ -0,0 +1,68 @@
...
+// 	}
+
+// 	rcv := conf.Receivers["x"].Receiver.(*receiver.Rabbitmq)
+
+// 	err = rcv.Start()
+
+// 	if err != nil {
+// 		t.Error(err)
+// 	}
+// }
```
sklirg: I prefer to remove this instead of committing a lot of commented-out code :-)

sender/rabbitmq.go
```diff
@@ -0,0 +1,136 @@
...
+	"time"
+
+	amqp "github.com/rabbitmq/amqp091-go"
+	"github.com/telenornms/skogul"
+	"github.com/telenornms/skogul/encoder"
+)
+
+type Rabbitmq struct {
+	Username string            `doc:"Username for rabbitmq instance"`
+	Password string            `doc:"Password for rabbitmq instance"`
```
sklirg: This can use the `Secret` type: https://github.com/telenornms/skogul/blob/primary/common.go#L348

sender/rabbitmq.go
```diff
@@ -0,0 +1,136 @@
...
+	Timeout  int               `doc:"Timeout for rabbitmq instance connection. Fallback is 10 seconds."`
+	channel  *amqp.Channel
+	once     sync.Once
+}
+
+func (r *Rabbitmq) init() {
+	if r.Username == "" || r.Password == "" {
+		fmt.Errorf("Error missing username or password")
+		return
+	}
```
sklirg: I would prefer to have this in a `Verify` function instead of in the `init` function. Verify is called when Skogul starts, and will therefore cause Skogul to exit with an error message immediately if this is misconfigured. If it is checked first in `init`, this will not trigger an error before the first message is attempted sent (which is when init() runs).

sender/rabbitmq.go
```diff
@@ -0,0 +1,136 @@
...
+		r.Queue,
+		false,
+		false,
+		false,
+		false,
+		nil,
+	)
+
+	if err != nil {
+		fmt.Errorf("Error %v", err)
```
sklirg: This should be logged (e.g. `log.WithError(err).Error("rabbitmq sender failed")`).
Furthermore, `fmt.Errorf()` returns an `error` type, it doesn't actually print anything, so this is just creating the `error` type and not assigning it/printing it 😅

sender/rabbitmq.go
```diff
@@ -0,0 +1,136 @@
...
+
+	conn, err := amqp.Dial(fmt.Sprintf("amqp://%s:%s@%s:%s/", r.Username, r.Password, r.Host, r.Port))
+	if err != nil {
+		fmt.Errorf("Failed initializing broker connection: %v", err)
+		return
+	}
+
+	ch, err := conn.Channel()
+	if err != nil {
+		fmt.Errorf("Error %v", err)
```
sklirg: This should be logged (e.g. `log.WithError(err).Error("rabbitmq sender failed")`).
Furthermore, `fmt.Errorf()` returns an `error` type, it doesn't actually print anything, so this is just creating the `error` type and not assigning it/printing it 😅

sender/rabbitmq.go
```diff
@@ -0,0 +1,136 @@
...
+		r.Timeout = 10
+	}
+
+	if r.Encoder.E == nil {
+		r.Encoder.E = encoder.JSON{}
+	}
+
+	conn, err := amqp.Dial(fmt.Sprintf("amqp://%s:%s@%s:%s/", r.Username, r.Password, r.Host, r.Port))
+	if err != nil {
+		fmt.Errorf("Failed initializing broker connection: %v", err)
```
sklirg: This should be logged (e.g. `log.WithError(err).Error("rabbitmq sender failed")`).
Furthermore, `fmt.Errorf()` returns an `error` type, it doesn't actually print anything, so this is just creating the `error` type and not assigning it/printing it 😅

sklirg (commented):
A couple of comments 🧹

sender/rabbitmq_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	r := Rabbitmq{
+		Username: "guest",
+		Password: "guest",
+		Queue:    "test-queue",
+	}
+
+	i := 0
+	for i < 100000 {
```
sklirg, line 86: This could be a typical for-loop instead of while-style:

```go
for i := 0; i <= 100000; i++ {
    ...
}
```

PR #312, opened 2025-01-15
Chore/whitespace

aslak01:
This PR contains mostly nits from the stricter [`gofumpt`](https://github.com/mvdan/gofumpt) go formatter about whitespace and error formatting without initial Uppercase and ending punctuation, and a couple changes from snake_case and SCREAMING_SNAKE_CASE to camelCase as it recommends. 

It also has some more concise variable declaration changes, probably in part due to new syntax since the code was originally written, and some stylistic `gofumpt` changes.

Nothing here should affect anything in the actual running of the code (with one exception which I will make a comment about), but in my opinion it makes the code more internally consistent and easier to read as a human consumer of it. 

Edit: Once committed, this PR's hash should be added to `.git-blame-ignore-revs` to still show who made the original code and not hide it with formatting changes.


transformer/unflatten.go
```diff
@@ -114,7 +112,7 @@ func (u *Unflatten) recursivelyCreateMap(root map[string]interface{}, keys []str
 	if _, ok := root[keys[pos]]; ok { // Key already exists, continue creating key-value pairs in the same map until we reach length of keys.
 		key, okk := root[keys[pos]].(map[string]interface{})
 		if !okk {
-			return root, errors.New(fmt.Sprintf("trying to write a value with type `%v` to a key without depth.", key))
+			return root, fmt.Errorf("trying to write a value with type `%v` to a key without depth", key)
```
aslak01, line 115: This has some potential runtime implications, the change is implemented according to the lsp/linter recommendations, but I'm not sure if there was a reason for the old way it was written.

PR #317, opened 2025-04-28
chore: remove macos tests from ci

aslak01:
The tests running on macos have become flaky due to internal github server issues causing these runtimes to be slow.
See e.g. https://github.com/actions/runner-images/issues/11509

Since skogul tests take execution speed into account this slowness causes random tests to fail without a good reason since we're not aiming to run skogul on macos environments in production.


kamilernerd:
lgtm

kamilernerd (approved):

